PR #24, opened 2014-04-10
add .npmignore file

dougwilson:
Hi! I've added a `.npmignore` file for you here that will remove the test folder from the npm package, making it slimmer since they are not used at that point. It helps those that check in their `node_modules` directory for server deploys :)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,2 @@
+test
+.travis.yml


AliChraghi:
> Hi! I've added a `.npmignore` file for you here that will remove the test folder from the npm package, making it slimmer since they are not used at that point. It helps those that check in their `node_modules` directory for server deploys :)

Your First Commit!

PR #32, opened 2015-05-28
Adds "First-Party-Only" attribute

victorenator:
Spec: https://tools.ietf.org/html/draft-west-first-party-cookies-03
Chromium: https://www.chromestatus.com/features/4672634709082112

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,9 @@ The serialize function takes a third parameter, an object, to set cookie options
 ### httpOnly
 > true or false
 
+### firstPartyOnly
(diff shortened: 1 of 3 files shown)


dougwilson:
Thanks! If you are willing, it would be nice to include a readme update + tests so we can accept it quicker :)

victorenator:
First-Party-Only: Tests and Readme

dougwilson:
Sorry for the crazy delay on merging this :S It has just been published to npm has version 0.2.0

PR #36, opened 2015-07-09
Adding more type safety checking an general code and jsdoc cleanup

arizonatribe:
* Some general cleanup (missing semi-colons, jsdoc return and param type inaccuracies, etc.). 
* Applied string param checking to `.serialize()` as was most recently added to `.parse()`. 
* Added more type safety checking against expected params inside the `options` param object.
* Made the maxAge validation more compact and protect against floating point numbers
* Modified `.parse()` to check for own property rather than check falsy when checking whether the params object already had a value set

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,7 +28,7 @@ var encode = encodeURIComponent;
  *
  * @param {string} str
  * @param {object} [options]
- * @return {string}
+ * @return {object}
  * @public
  */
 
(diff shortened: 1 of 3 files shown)


dougwilson:
Hi! Would you be willing to split this up into multiple PRs that have specific purposes so we can accept them at different rates? Right now I can review, but any issue will hold up accepting everything else that has no issue. It's up to you on how quickly you are looking to get any changes accepted is all :)

arizonatribe:
Hi, I'm not in any kind of hurry, but I'll break them out into smaller PRs and re-submit those individually.

Thanks.

PR #37, opened 2015-07-10
General cleanup of jsdoc, adding missing semi-colons and unit tests

arizonatribe:
Some general cleanup (missing semi-colons, jsdoc return and param type inaccuracies) and added more test coverage.

### DIFF
--- a/test/parse.js
+++ b/test/parse.js
@@ -53,4 +53,13 @@ test('dates', function() {
             cookie.parse('priority=true; expires=Wed, 29 Jan 2014 17:43:25 GMT; Path=/',{
                 decode: function(value) { return value; }
             }));
-})
+});
+
+test('assign only once', function() {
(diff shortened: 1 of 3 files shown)


test/parse.js
```diff
@@ -53,4 +53,13 @@ test('dates', function() {
             cookie.parse('priority=true; expires=Wed, 29 Jan 2014 17:43:25 GMT; Path=/',{
                 decode: function(value) { return value; }
             }));
-})
+});
+
+test('assign only once', function() {
+    assert.deepEqual({ foo: '%1', bar: 'bar' },
+        cookie.parse('foo=%1;bar=bar;foo=boo;HttpOnly;Secure'));
```
dougwilson: Please remove the `HttpOnly;Secure` part from these strings, as that's not part of the `Cookie` header.

arizonatribe:
Removed the properties as per your request

PR #38, opened 2015-07-10
Added string parameter validation to serialize() method same as is in…

arizonatribe:
Applied string param checking to `.serialize()` as was most recently added to `.parse()`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -84,6 +84,13 @@ function parse(str, options) {
  */
 
 function serialize(name, val, options) {
+  if (typeof name !== 'string') {
+    throw new TypeError('argument name must be a string');
+  }
+  if (typeof val !== 'string') {
+    throw new TypeError('argument val must be a string');
+  }
+
   var opt = options || {};
   var enc = opt.encode || encode;
   var pairs = [name + '=' + enc(val)];
--- a/test/parse.js
+++ b/test/parse.js
@@ -6,8 +6,21 @@ var cookie = require('..');
 suite('parse');
 
 test('argument validation', function() {
-    assert.throws(cookie.parse.bind(), /argument str must be a string/);
-    assert.throws(cookie.parse.bind(null, 42), /argument str must be a string/);
+  var errMessage = /argument str must be a string/;
+
+  assert.throws(cookie.parse.bind(), errMessage);
+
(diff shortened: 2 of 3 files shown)


test/parse.js
```diff
@@ -6,8 +6,11 @@ var cookie = require('..');
 suite('parse');
 
 test('argument validation', function() {
+    var nonStrings = [1010, true, {}, (function() {return;}), new Date(), []];
```
dougwilson: Please just do one `assert.throws` line for each instead of iterating an array; an error won't tell you which value was allowed though, only point to line 12, which is not helpful.

test/serialize.js
```diff
@@ -5,6 +5,15 @@ var cookie = require('..');
 
 suite('serialize');
 
+test('argument validation', function() {
+    var nonStrings = [1010, true, {}, (function() {return;}), new Date(), []];
```
dougwilson: Please just do one `assert.throws` line for each instead of iterating an array; an error won't tell you which value was allowed though, only point to line 12 or 13, which is not helpful.

arizonatribe:
Modified the specs as per your requests.

index.js
```diff
@@ -84,6 +84,13 @@ function parse(str, options) {
  */
 
 function serialize(name, val, options) {
+  if (typeof name !== 'string') {
+    throw new TypeError('argument name must be a string');
+  }
+  if (typeof val !== 'string') {
```
dougwilson: I have to think on this new restriction a little bit, because it's a breaking change since it currently works just fine with non-strings (especially numbers), which people are likely to be providing currently.

arizonatribe: Would it be enough to skip the verifying that `val` is a string when the `encode` property has been defined for the `options` object? That would allow the custom encoder the user defines to handle `val` however they see fit.

dougwilson: The current encoder works just fine with numbers, Dates, etc. so it would still be a breaking change.

dougwilson:
Looks like the source repo is gone, so I wasn't able to rebase and merge into the 1.0 branch. If you would like to see the validation, please feel free to open a new PR :) ! Sorry our 1.0 release languashed for so long.

PR #39, opened 2015-07-10
Improved the type validation for max-age to guard against floating po…

arizonatribe:
Prevents floating point values for max-age, parses as int yet allows for string int values (including zero).

### DIFF
--- a/index.js
+++ b/index.js
@@ -88,12 +88,8 @@ function serialize(name, val, options) {
   var enc = opt.encode || encode;
   var pairs = [name + '=' + enc(val)];
 
-  if (null != opt.maxAge) {
-    var maxAge = opt.maxAge - 0;
-    if (isNaN(maxAge)) throw new Error('maxAge should be a Number');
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -88,12 +88,8 @@ function serialize(name, val, options) {
   var enc = opt.encode || encode;
   var pairs = [name + '=' + enc(val)];
 
-  if (null != opt.maxAge) {
-    var maxAge = opt.maxAge - 0;
-    if (isNaN(maxAge)) throw new Error('maxAge should be a Number');
-    pairs.push('Max-Age=' + maxAge);
-  }
-
+  (typeof opt.maxAge === 'number' || opt.maxAge === '0' || ~~opt.maxAge !== 0)
```
dougwilson: Please make this an actual `if` statement.

arizonatribe: Modified to be an `if` statement as per your request

dougwilson:
Merged. Thanks, again!

PR #40, opened 2015-07-10
Slight improvements to options object member validation

arizonatribe:
Explicitly checking for options object members expected data types (`typeof 'function'` for encode/decode, valid `Date` or date strings for `expiration`, valid `string` values for `domain` and `path`), also using shorthand for "if value is acceptable add to array" logic.

### DIFF
--- a/index.js
+++ b/index.js
@@ -38,9 +38,9 @@ function parse(str, options) {
   }
 
   var obj = {}
-  var opt = options || {};
+  var opt = typeof options === 'object' ? options : {};
   var pairs = str.split(/; */);
-  var dec = opt.decode || decode;
+  var dec = typeof opt.decode !== 'undefined' ? opt.decode : decode;
 
   pairs.forEach(function(pair) {
     var eq_idx = pair.indexOf('=')
@@ -84,8 +84,11 @@ function parse(str, options) {
  */
 
 function serialize(name, val, options) {
-  var opt = options || {};
-  var enc = opt.encode || encode;
+  var opt = typeof options === 'object' ? options : {};
+  var enc = typeof opt.encode !== 'undefined' ? opt.encode : encode;
+  if (typeof enc !== 'function') {
+    throw new TypeError('encode option must be a function');
+  }
   var pairs = [name + '=' + enc(val)];
 
   if (null != opt.maxAge) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -94,11 +94,12 @@ function serialize(name, val, options) {
     pairs.push('Max-Age=' + maxAge);
   }
 
-  if (opt.domain) pairs.push('Domain=' + opt.domain);
-  if (opt.path) pairs.push('Path=' + opt.path);
-  if (opt.expires) pairs.push('Expires=' + opt.expires.toUTCString());
-  if (opt.httpOnly) pairs.push('HttpOnly');
-  if (opt.secure) pairs.push('Secure');
+  typeof opt.domain === 'string' && pairs.push('Domain=' + opt.domain);
```
dougwilson: Please undo these changes so they are actual `if` statements.

index.js
```diff
@@ -84,8 +84,8 @@ function parse(str, options) {
  */
 
 function serialize(name, val, options) {
-  var opt = options || {};
-  var enc = opt.encode || encode;
+  var opt = typeof options === 'object' ? options : {};
+  var enc = typeof opt.encode === 'function' ? opt.encode : encode;
```
dougwilson: There should be an error given `{encode: 42}` since it's not a function.

index.js
arizonatribe: Modified to be `if` statements as per your request

index.js
arizonatribe: Added in spec to verify fall-through to default `encode` function when `{encode: 42}`

dougwilson: Right, but fall-through is not good behavior; it should error. It errored before, though with a not great error, but it should still error.

arizonatribe: Actually, it didn't always error before. Any falsy value for `opt.encode` would fall through to the default `encode` function. Maybe the intent of the original logical OR was to see if the "encode" property was defined on the options object or not, if that's the case, what about using `typeof opt.encode !== 'undefined' ? opt.encode : encode`. That should allow any defined value for `opt.encode` which isn't a function to error out.

arizonatribe: Modified it to apply those changes (only falls through to the default if the "encode" property hasn't been defined for the `options` object.

test/serialize.js
```diff
@@ -61,6 +131,14 @@ test('parse->serialize', function() {
       cookie.serialize('cat', ' ";/')));
 });
 
+test('encoded method validation', function() {
+   assert.throws(cookie.serialize.bind(null, 'cat', ' ', {
+       encode: 42
+   }), /number is not a function/);
```
dougwilson: Looks like this is failing on io.js 2.x due to a difference in the v8 error message.

arizonatribe: I saw that earlier and threw in a regex that should cover it in either case. Let me know though.

dougwilson: Why not just throw our own error, so the message is predictable (and probably more helpful to users)?

arizonatribe: Okay, added code to throw after parsing encode

dougwilson:
Sorry @arizonatribe that I had neglected this pull request. I have cherry-picked the couple checks that were definitely backwards-compatible so we can use at least something from the pull request (it's not always easy to accept a PR when it does several different things, and some have issues while other changes do not). Please feel free to make a new PR adding any additional checks you would like and we can get it into the 1.0 coming up soon!

PR #42, opened 2015-08-13
cookie.expires can be

carlosascari:
* a Date object
* a unix timestamp
* a date string with ISO 8601 format
* a date string with RFC2822 / IETF date syntax

### DIFF
--- a/index.js
+++ b/index.js
@@ -96,7 +96,7 @@ function serialize(name, val, options) {
 
   if (opt.domain) pairs.push('Domain=' + opt.domain);
   if (opt.path) pairs.push('Path=' + opt.path);
-  if (opt.expires) pairs.push('Expires=' + opt.expires.toUTCString());
+  if (opt.expires) pairs.push('Expires=' + new Date(opt.expires).toUTCString());
   if (opt.httpOnly) pairs.push('HttpOnly');
   if (opt.secure) pairs.push('Secure');
 
--- a/index.js
+++ b/index.js
@@ -96,7 +96,9 @@ function serialize(name, val, options) {
 
   if (opt.domain) pairs.push('Domain=' + opt.domain);
   if (opt.path) pairs.push('Path=' + opt.path);
-  if (opt.expires) pairs.push('Expires=' + new Date(opt.expires).toUTCString());
+  var now = new Date(opt.expires);
+  var utc = new Date(now.getTime() + now.getTimezoneOffset() * 60 * 1000);
+  if (opt.expires) pairs.push('Expires=' + utc.toUTCString());
   if (opt.httpOnly) pairs.push('HttpOnly');
   if (opt.secure) pairs.push('Secure');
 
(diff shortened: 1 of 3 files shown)


dougwilson:
Awesome, thank you! It seems the tests are currently failing on Travis CI, if you can take a look. Also, with this change, what happens if someone provide an invalid string? for example `{ expires: 'foo' }`?

carlosascari:
Hey, if you provide an invalid string or some other non-falsy, invalid value, it will throw. I've added a test for this.

index.js
```diff
@@ -96,7 +96,13 @@ function serialize(name, val, options) {
 
   if (opt.domain) pairs.push('Domain=' + opt.domain);
   if (opt.path) pairs.push('Path=' + opt.path);
-  if (opt.expires) pairs.push('Expires=' + opt.expires.toUTCString());
+  if (opt.expires) 
+  {
+    var now = new Date(opt.expires);
+    if (now + '' === 'Invalid Date') throw new Error('expires should be a valid datestring, unix timestamp or Date object');
+    var utc = new Date(now.getTime() - now.getTimezoneOffset() * 60 * 1000);
```
dougwilson: I have no idea what you're trying to accomplish here, but it produces the wrong date/time:
```js
var str = '03 08 2015 07:10:00 -0400';

console.log(new Date(str).toISOString()); // correct time: 2015-03-08T11:10:00.000Z

var now = new Date(str);
var utc = new Date(now.getTime() - now.getTimezoneOffset() * 60 * 1000);

console.log(utc.toISOString()); // wrong time: 2015-03-08T11:10:00.000Z
```

dougwilson: I have no idea what you're trying to accomplish here, but it produces the wrong date/time:
```js
var str = '03 08 2015 07:10:00 -0400';

console.log(new Date(str).toISOString()); // correct time: 2015-03-08T11:10:00.000Z

var now = new Date(str);
var utc = new Date(now.getTime() - now.getTimezoneOffset() * 60 * 1000);

console.log(utc.toISOString()); // wrong time: 2015-03-08T07:10:00.000Z
```

carlosascari:
You're right I should have explained.

The Date object does not have native support for timezones. Simple fixes seem to be [unreliable] (http://stackoverflow.com/questions/948532/how-do-you-convert-a-javascript-date-to-utc) in general. 
You might have noticed, on the previous build, it passed the mocha tests on certain versions of node.js, but not on the rest because the datestring was off by a few hours.

I used the solution proposed by [Gras Double](http://stackoverflow.com/a/11964609), to normalize the timezone as UTC. The solution worked on my side and on the Travis servers, so i left it in good faith, my mistake.

The only way to have a reliable conversion, as I understand it is by setting a timezone to determine the offset: such as "America/New_York" **-05:00**, so as not to rely on the client's timezone,

Correct me if i'm wrong

dougwilson:
Hi @carlosascari , I'm going to close this PR because it cannot be merged any more and it has the problem above about how it creates the wrong date/times. Please feel free to open a new PR with the changes rebased and the issues above resolved :)

Issue #45, opened 2015-09-24
.parse missing HttpOnly and Secure


emilbayes:
Why are the HttpOnly and Secure flags missing? ((See test))[https://github.com/jshttp/cookie/blob/master/test/parse.js#L36]

I would have expected:

```js
{ foo: '%1', bar: 'bar', Secure: true, HttpOnly: true}
```

I haven't read the RFC so I don't know the details, but those two flags are useful when building a HTTP Agent

hsablonniere:
Same question?

dougwilson:
Hi! You are probably confusing the `Cookie` header and the `Set-Cookie` header; the `.parse` method of this module only parses the `Cookie` header. That example is just sitting there probably confusing you, as the `Cookie` header does not have that. The test may as well be `cookie.parse('foo=%1;bar=bar;hi;bye')`. If you would like to have a method to parse the `Set-Cookie` header, please, feel free to make a PR to implement this functionality :)

PR #46, opened 2015-09-24
Parse Set-Cookie HttpOnly and Secure with boolean true value

hsablonniere:
I understand your explanation in #45 but I'm using your lib to test an API and I wanted to verify the Set-Cookie header from a response so I needed this simple fix.

Maybe we need something more elaborate to parse Cookie header and Set-Cookie header differently.

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,30 +49,36 @@ function parse(str, options) {
     throw new TypeError('argument str must be a string');
   }
 
-  var obj = {}
+  var obj = {};
   var opt = options || {};
   var pairs = str.split(pairSplitRegExp);
   var dec = opt.decode || decode;
 
   pairs.forEach(function(pair) {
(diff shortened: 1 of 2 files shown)


dougwilson:
This will break people using .parse for the Cookie header, so unfortunately I cannot accept this as-is. Please implement a proper Set-Cookie header parser, including it being a different API so people can use the right one.

Altering a test indicates your change is a breaking API change, and there is no timeline when we will release a breaking API change, especially when it doesn't really make sense.

dougwilson:
If you need something ASAP, it looks like the module https://github.com/juji/simple-cookie is made to parse the `Set-Cookie` header, which should do exactly what you're looking for.

hsablonniere:
Yeah, no worries, I totally agree. Thanks for the simple to simple-cookie. It's just what I needed ;-)

PR #47, opened 2015-10-25
Serialize objects of key/value pairs

### LINKED ISSUE #30: Update serialize to accept a object of key/values.
The serialize method should be updated to accept an object of key/values and serialize to a cookie string. This would make the serialize/parse methods complement each other.

gatkin:
Updates the serialize method to accept objects of key-value pairs. ~~If the object of key/value pairs contains multiple pairs, then allow the optional options object to specify only an encoding. If the object of key/value pairs contains only one pair, then the options object may specify cookie parameters as well as an encoding.~~ Resolves #30

### DIFF
--- a/index.js
+++ b/index.js
@@ -79,8 +79,6 @@ function parse(str, options) {
   return obj;
 }
 
-
-
 /**
  * Serialize data into a cookie header.
  *
@@ -107,6 +105,7 @@ function parse(str, options) {
  * @return {string}
  * @public
  */
+
 function serialize(name, val, options) {
   if( typeof name === 'object') {
     var cookies = name;
@@ -135,7 +134,6 @@ function serialize(name, val, options) {
   }
 }
 
-
 /**
  * Serialize name value pair into a cookie header.
  *
@@ -151,6 +149,7 @@ function serialize(name, val, options) {
  * @return {string}
  * @private
  */
+
 function serializeNameValue(name, val, options) {
   var opt = options || {};
   var enc = opt.encode || encode;
--- a/index.js
+++ b/index.js
@@ -83,11 +83,9 @@ function parse(str, options) {
  * Serialize data into a cookie header.
  *
  * If the first parameter is an object, serialize the key-value pairs
- * in the object into a cookie string suitable for http headers. An
- * optional options object can be used to specify the encoding. If only
- * one key-value pair is in the first parmater, then the options object can also
- * specify cookie parameters. If more than one key-value pairs are in the
- * first parameter, then the options object may only specify an encoding.
+ * in the object into an array of cookie strings suitable for http headers.
+ * An optional options object can be used to specify the encoding and cookie
+ * parameters.
  *
  * If the first parameter is a string, serialize the name value pair
  * into a cookie string suitable for http headers. An optional options
@@ -96,8 +94,8 @@ function parse(str, options) {
  * serialize('foo', 'bar', { httpOnly: true })
  *   => "foo=bar; httpOnly"
  *
- * serialize({ foo: 'bar', cat: 'meow' })
- *  => "foo=bar; cat=meow"
+ * serialize({ foo: 'bar', cat: 'meow' }, { httpOnly: true })
+ *  => ["foo=bar; httpOnly", "cat=meow; httpOnly"]
  *
  * @param {string} name
  * @param {string} val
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -96,6 +107,50 @@ function parse(str, options) {
  */
 
 function serialize(name, val, options) {
+  if( typeof name === 'object') {
+    var cookies = name;
+    var serializeOptions = val;
+
+    var cookieNames = Object.keys(cookies);
+    if(0 === cookieNames.length) {
+      return '';
```
dougwilson: Perhaps return `undefined` for this case? Not too clear, but just feel like setting the `Set-Cookie` header to the result would make more sense if `undefined` was returned, since that would just not send the header instead of send an empty header.

index.js
```diff
@@ -96,6 +107,50 @@ function parse(str, options) {
...
+    } else if(cookieNames.length > 1) {
```
dougwilson: Why the `else if` case here? There is some case there `Object.keys` would return negative numbers?

index.js
```diff
@@ -96,6 +107,50 @@ function parse(str, options) {
...
+      // If there are more than one cookies to serialize, only allow
+      // an encoding option to be set
+      var opt = serializeOptions || {};
+      serializeOptions = {
```
dougwilson: This doesn't really make any sense. People supplying the object as a variable would have no way to understand if their options are going to be honored. Besides, it seems perfectly valid to want to set all the cookies to have `httpOnly: true`, `secure: true`, etc.

index.js
```diff
@@ -96,6 +107,50 @@ function parse(str, options) {
...
+        encode: opt.encode || encode
+      };
+    }
+
+    var serializedCookies = [];
+    cookieNames.forEach(function(cookieName) {
+        serializedCookies.push(serializeNameValue(cookieName, cookies[cookieName], serializeOptions));
```
dougwilson: We already know the number of cookies ahead time time; forcing the runtime to keep expanding the array is not very efficient.

index.js
gatkin: I could be wrong about this, but my thinking was that if we are given several cookies to serialize, they want those cookies in the `Cookie` request header which, from what I read in [RFC6265 Section 5.4](http://tools.ietf.org/html/rfc6265#section-5.4) should be serialized as the cookie name followed by an '=' followed by the cookie value with all cookies separated by a semicolon. So if we want to serialize multiple cookies into a single string, we should not allow additional cookie parameters in the string (again, I could be wrong about this, please correct me if so).

One possible option is: given an object with multiple cookies and an options object with cookie parameters we can return an array of serialized cookie strings, one for each cookie, with all the parameters set as could be used in a `Set-Cookie` header. Another option is to let the caller indicate in the options object whether they want their cookies serialized for a `Set-Cookie` response header or a `Cookie` request header.

dougwilson: The `serialize` method outputs a string that is for the `Set-Cookie` response header, not the `Cookie` request header.

gatkin: I see, so if we are given multiple cookies to serialize, would it be best to return an array of serialized cookie strings rather than a single string with all cookies?

dougwilson: Yea, that makes sense :)

index.js
```diff
@@ -96,6 +105,43 @@ function parse(str, options) {
...
+    var cookieNames = Object.keys(cookies);
+    if(0 === cookieNames.length) {
+      return undefined;
+    } else {
+      var serializedCookies = new Array(cookieNames.length);
+      for(var i=0; i<cookieNames.length; i++) {
+          serializedCookies[i] = serializeNameValue(cookieNames[i], cookies[cookieNames[i]], serializeOptions);
+      }
+
+      return serializedCookies;
```
gatkin: Even if given a single key-value pair in the object, this returns an array of serialized cookie strings. I thought this might make the function more consistent in that if you pass an object, you always get back an array of strings rather than returning just a single string if you give only a single key-value pair in the object and an array if you give multiple key-value pairs in the object.

I still left the behavior that when given a name string and a value string, the function returns a single serialized cookie string so this does not break anyone's code that uses the old version of this function.

gatkin:
What would you think about having separate methods for parsing/serializing the Set-Cookie response header and the Cookie request header? Something like `parseSetCookie`, `serializeSetCookie`, `parseCookieRequest`, `serializeCookieRequest`? If this is a worthwhile change, I can make an issue and work on a pr for this.

anthonator:
What's left to do on this?

travisperson:
Ping

This would be really great to have.

frogcjn:
any updates？

jgornick:
To follow up on this, my workaround was to do something like:

```
const cookies = { foo: 'bar', cat: 'meow', dog: 'ruff' };
const serializedCookies = Object.keys(cookies).map(key => cookie.serialize(key, cookies[key]);
```

The `serializedCookies` is now an array of serialized key/value pairs (as string).

The single line of using `Object.keys` and `Array.map` made me think if there's really a need to overload the `serialize` method with this ability.

If anything, instead of overloading the method, I would maybe add a `serializeObject` method with a type of `serializeObject :: Object -> Array`

TomVance:
Was looking around for something like this today, happy to take this over if it's something still of interest to people?

export-mike:
@TomVance @jgornick @NedalEldeen @gatkin @dougwilson 

I have wanted this feature recently and updated a new PR against the current state of the main branch.

https://github.com/jshttp/cookie/pull/119

dougwilson:
Hi @gatkin I know this is old, but I'm trying to rebase and merge your changes. It seems I don't have the ability to push the rebase to your branch. If you're still around and can check the box on the right-hand side in this PR to allow me to push, that would be awesome. If not, no problem. I'll still get the changes landed, they just won't appear as the PR is merged.

blakeembrey:
I'm going to close this out as won't fix for now, but if there's strong demand for this I can see adding it as a new method to the library. For now, I think the need to have different options per cookie and the simplicity of just calling serialize twice or using your own wrappers is better than making the library more complex.

PR #48, opened 2015-10-26
Add script to make minifid umd bundle with source map

TheLudd:
Adds two npm scripts `browserify` and `minify` that compiles the code
into an umd format, minified and with source maps into the folder `dist`
that is ignored by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 coverage/
+dist/
 node_modules/
 npm-debug.log
--- a/package.json
+++ b/package.json
@@ -13,19 +13,24 @@
   ],
   "repository": "jshttp/cookie",
   "devDependencies": {
+    "browserify": "11.2.0",
     "istanbul": "0.3.22",
-    "mocha": "1.21.5"
+    "mocha": "1.21.5",
+    "uglify-js": "2.5.0"
   },
   "files": [
     "HISTORY.md",
     "LICENSE",
     "README.md",
+    "dist/",
     "index.js"
   ],
   "engines": {
     "node": ">= 0.6"
   },
   "scripts": {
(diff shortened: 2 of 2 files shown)


TheLudd:
I'd be glad if this was accepted and also if it was included in the release process. I was not sure how to do that, and I could add it to this PR if instructed how to.

dougwilson:
I have never used browserify or anything and so have no idea. Unless it's part of a process where I cannot accidentally forget to perform it, then I cannot accept the PR, as it is a divergence from every other project.

dougwilson:
P.S. it looks like the Travis CI integration is broken. The build for your PR (https://travis-ci.org/jshttp/cookie/builds/87499386) shows this PR breaks our tests on 0.6 and 0.8, two supported platforms.

TheLudd:
How is this published? Do you just run `npm publish` or do you do anything else?

TheLudd:
The build failures seems to be due to the depedencies specified with the "^" prefix which does not seem to be supported in node 0.6 and 0.8. 

I removed it from the dependencies I added but it still fails due to sub dependencies having the prefix. However I also found that Istanbul has the same type of dependencies and it works. Do you know anything about this?

dougwilson:
> How is this published? Do you just run npm publish or do you do anything else?

I run `npm pack`, inspect the tarball to verify it looks right and then publish that specific tarball to npm.

This whole time I thought you were proposing to have the minified source code committed to git to pull down from some other package manager. I now see it's actually to bloat our npm package. We specifically work to ensure our installed npm package is lean & does not contain any duplicate code.

Because of this reason and that something like `browserify` can do this same thing with an installed npm module, we're going to close this PR, I'm sorry.

PR #52, opened 2016-02-01
Test node stable

revington:
I added `node` to the matrix of versions. `node` is a pointer to latest node, actually 5.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ node_js:
     - "3.3"
     - "4.0"
     - "4.2"
+    - "node"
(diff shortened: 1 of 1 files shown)


dougwilson:
Thanks for this pull request! We actually specific specific minor versions to test since historically we end up in a situation where a new version comes out that breaks our tests in some way, even if it's not because Node.js issues. For example, Node.js 5.3+ is still shipping with a broken `npm` that does not allow removing non-installed modules, which breaks testing in various modules of our organization still. I added 5.6 to the testing matrix.

Issue #55, opened 2016-04-17
Add support for "SameSite" attribute


lpfeup:
"Same-site cookies allow servers to mitigate the risk of CSRF and information leakage attacks by asserting that a particular cookie should only be sent with requests initiated from the same registrable domain."

https://tools.ietf.org/html/draft-west-first-party-cookies-06
https://www.chromestatus.com/feature/4672634709082112

dougwilson:
We can probably work to add support for this attribute, sure! Before we start, can you explain what the difference between this attribute and the first-party-only one is? The abstract in the RFC drafts at least seem to be very similar.

jltwoo:
It's just renaming, it's the same attribute

They renamed it due to legal reasons, because "party" is a legal term used for drafting laws... so "SameSite" was recommended instead to prevent confusion.
https://lists.w3.org/Archives/Public/ietf-http-wg/2016JanMar/0134.html

dougwilson:
@jltwoo, ah, gotcha. So that asks the question for what we should do with the existing `firstPartyOnly` property. I can think of three options, and I'm not sure which is best:

1. Do nothing and have both of them.
2. Mark `firstPartyOnly` as deprecated and tell users to use `sameSite`, but take no actions.
3. Mark `firstPartyOnly` as deprecated, tell users to use `sameSite` and also set the same site flag.

Thoughts?

wesleytodd:
I did some googling and couldn't find much information on the browser support for these to options.  Do you have that info @jltwoo?  If the browsers take either or none then I am in favor of 3 to get people using the correct flag.

This is the closes thing I could find that actually talks about how this will move forward: https://www.chromestatus.com/features/4672634709082112

dougwilson:
So, circling back around on this issue, I did some digging around, and it looks like Chrome never actually shipped `First-Party-Only`, at least without requiring a flag set in the client. This means it is unlikely to actually have ever been accepted anywhere. With this module still at the `0.x` phase in semver, I move to just add same site, and simply drop first party only (which means supplying it in the option in version 0.3.0+ will just do nothing, as if the option didn't exist). Thoughts?

lpfeup:
I am also in favor of deprecating `firstPartyOnly ` and just move to `sameSite`.

jltwoo:
I agree with @dougwilson, if you follow the full thread here: 
https://bugs.chromium.org/p/chromium/issues/detail?id=459154

they never actually shipped it until Chrome 51
[...]

PR #56, opened 2016-04-17
Adds "SameSite" attribute.

lpfeup:
#55

### DIFF
--- a/index.js
+++ b/index.js
@@ -133,6 +133,19 @@ function serialize(name, val, options) {
     pairs.push('Path=' + opt.path);
   }
 
+  if (null != opt.sameSite) {
+    if(opt.sameSite !== true && opt.sameSite !== false && opt.sameSite !== 'Strict' && opt.sameSite !== 'Lax'){
+      throw new TypeError('option sameSite is invalid');
+    }
+
+    if(opt.sameSite === true) {
+      pairs.push('SameSite');
+    }
+    else if(opt.sameSite === 'Strict' || opt.sameSite === 'Lax') {
+      pairs.push('SameSite=' + opt.sameSite);
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


dougwilson:
Hi @lpfeup, thanks for the pull request! There is an on going discussion in #55 with open questions that is holding up the acceptance of this pull request. If you could participate in them, that would be awesome :)

index.js
```diff
@@ -133,6 +133,19 @@ function serialize(name, val, options) {
     pairs.push('Path=' + opt.path);
   }
 
+  if (null != opt.sameSite) {
+    if(opt.sameSite !== true && opt.sameSite !== false && opt.sameSite !== 'Strict' && opt.sameSite !== 'Lax'){
+      throw new TypeError('option sameSite is invalid');
```
dougwilson: Reading the spec, the value is not case-sensitive, so we should accept `strict`, `lax` or any other case combination. Please update this to match the spec :)

lpfeup:
@dougwilson I added support for case-insensitive values to match the spec.

index.js
```diff
@@ -136,6 +136,21 @@ function serialize(name, val, options) {
     str += '; Path=' + opt.path;
   }
 
+  if (null != opt.sameSite) {
+    var sameSiteRegExp = new RegExp(/^(strict|lax)$/i);
```
dougwilson: `new RegExp` is not necessary, you can just use a literal. Also, please hoist the expression up with the other expressions, since v8 is very slow when compiling expressions.

dougwilson:
Thanks so much, @lpfeup! I made a couple additional tweaks from some profiling (and to normalize the value of the attribute, no matter the incoming casing), I hope you don't mind! You are marked as the author of the merged & tweaked commit.

PR #69, opened 2017-08-23
Clean up switch statement

bdsomer:


dougwilson:
Hi @bdsomer thanks for the pull request! The style adhered to here is the "no fallthrough" rule (https://eslint.org/docs/rules/no-fallthrough) which is why the section is duplicated. Thanks anyway!

bdsomer:
@dougwilson ah, gotcha!

Issue #60, opened 2016-09-05
Reading duplicated cookies


a-x-:
my bro sends weird cookies (after deleting and installing cookies again):
```
user_id=; session_id=; session_id=XXXXXXX; user_id=6
```

`cookie-parser` parse them as a `''`

but expected reading a second value of each cookie

https://github.com/expressjs/cookie-parser/issues/31

dougwilson:
Hi @a-x-, that is correct, since the interface of this module only provides one value per name, so the first occurrence is chosen. I guess we can either completely change the interface to allow you to get all the values for a given name or we can change which one we can get if you can provide a link to some specification that says which occurrence we should actually be reading (the first or the last in your example).

It is perfectly legal for a web browser to send a `Cookie` header with multiple occurrences of the same cookie name, because a browser can have multiple overlapping cookies for the same page with the same name, differing on some attribute like `path`, `httpOnly`, or similar, so it isn't clear if you have two over-lapping cookies which to use.

Having a cookie with an empty value is a perfectly value cookie, so simply skipping over those a preferring ones with values is not on option, unless you know of a specification that says servers should be doing this.

a-x-:
oh, thank you

it's a kind of shit:
```
Although cookies are serialized linearly in the Cookie header,
   servers SHOULD NOT rely upon the serialization order.  In particular,
   if the Cookie header contains two cookies with the same name (e.g.,
   that were set with different Path or Domain attributes), servers
   SHOULD NOT rely upon the order in which these cookies appear in the
   header.
```

> servers   SHOULD NOT rely upon the order in which these cookies appear in the   header

https://tools.ietf.org/html/rfc6265#section-4.2.2

So, there i see the **solution**.

Can you add into your cookie parser conf same keys order policy?

e.g.: `{ sameKeysPolicy: 'first' | 'last' | 'collection' }`

a-x-:
You can use the `{ sameKeysPolicy: 'first' }` for compatibility reasons.
And, as I think, you should recommend using the `{ sameKeysPolicy: 'last' }` for new users

dougwilson:
Cool, feel fee to make a PR. If you at foing to recommend "last" in the documentation, you'll probably need to explain why one should choose "last" over "first", especially since that spec says do not rely on order, so it says doing either is equally bad, neither is recommended.

Perhaps it would be better to rewrite the interface to allow retrieval of all the cookies?

demurgos:
Hi,
[...]

PR #85, opened 2018-11-27
Add parseRepeated function

BernabeFelix:
This is a proposed solution for this #60 
It basically uses the same code than parse but instead of assign in once, it concatenates it to an array.

E.g.
```
'foo=false;bar=bar;foo=true' 

Becomes

{ foo: ['false', 'true'], bar: ['bar'] }
```

I tried to follow the current code style, used the same tests and benchmarks but I'm sure there's place for improvements.

@dougwilson

### DIFF
--- a/benchmark/parseRepeated.js
+++ b/benchmark/parseRepeated.js
@@ -1,0 +1,74 @@
+/**
+ * Module dependencies.
+ */
+
+var benchmark = require('benchmark')
+var benchmarks = require('beautify-benchmark')
+
+/**
+ * Globals for benchmark.js
+ */
+
+global.cookie = require('..')
+
+var suite = new benchmark.Suite()
+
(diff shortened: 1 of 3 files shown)


Poyoman39:
Any news about this PR ? Have the same issue with npm cookie

jacklrs:
Maybe this code duplication prevents this being merged. What about leverage existing "options" object adding a new key to configure desired behaviour? For instance it the name of this option could be "parseToArray" with default false ...

dougwilson:
Yes, @jacklrs that makes the most sense -- add an option to the existing `parse` function. It would have minimal code duplication as well.

dougwilson:
Actually, looks like this was implemented in #108 , so I'm going to close this in favor of that PR 👍

PR #88, opened 2019-02-15
build: implement using ES-modules syntax

### LINKED ISSUE #87: Consider to support ES "module"
It would be great to implement ES module entrypoint

jorgecasar:
Close #87

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ coverage/
 node_modules/
 npm-debug.log
 package-lock.json
+pkg
--- a/package.json
+++ b/package.json
@@ -13,6 +13,11 @@
   ],
   "repository": "jshttp/cookie",
   "devDependencies": {
+    "@pika/pack": "^0.3.2",
+    "@pika/plugin-build-node": "^0.3.11",
+    "@pika/plugin-build-web": "^0.3.11",
+    "@pika/plugin-standard-pkg": "^0.3.11",
+    "@pika/plugin-build-types": "^0.3.11",
     "beautify-benchmark": "0.2.4",
     "benchmark": "2.1.4",
     "eslint": "4.19.1",
@@ -24,7 +29,7 @@
     "HISTORY.md",
     "LICENSE",
     "README.md",
-    "index.js"
+    "src/index.js"
   ],
   "engines": {
     "node": ">= 0.6"
@@ -34,6 +39,15 @@
     "lint": "eslint --plugin markdown --ext js,md .",
     "test": "mocha --reporter spec --bail --check-leaks test/",
     "test-ci": "istanbul cover node_modules/mocha/bin/_mocha --report lcovonly -- --reporter spec --check-leaks test/",
-    "test-cov": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter dot --check-leaks test/"
+    "test-cov": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter dot --check-leaks test/",
+    "build": "pack build"
+  },
+  "@pika/pack": {
+    "pipeline": [
+      ["@pika/plugin-standard-pkg"],
+      ["@pika/plugin-build-node"],
+      ["@pika/plugin-build-web"],
+      ["@pika/plugin-build-types"]
+    ]
   }
 }
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 coverage
 node_modules
+pkg
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,3 +1,6 @@
+parserOptions:
+  ecmaVersion: 6
+  sourceType: module
 rules:
   eol-last: error
   indent: ["error", 2, { "SwitchCase": 1 }]
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,8 @@ before_install:
       npm prune
       npm rebuild
     fi
-
+before_script:
+  - npm run build
 script:
   - |
     # Run test script, depending on istanbul install
--- a/README.md
+++ b/README.md
@@ -16,8 +16,22 @@ $ npm install cookie
 
 ## API
 
+### CommonJS syntax
+
 ```js
 var cookie = require('cookie');
+```
+
+### ES module syntax
+
+```js
+import * as cookie from 'cookie';
+```
+
+or
+
+```js
+import { serialize, parse } from 'cookie';
 ```
 
 ### cookie.parse(str, options)
@@ -180,6 +194,12 @@ function onRequest(req, res) {
 }
 
 http.createServer(onRequest).listen(3000);
+```
+
+## Building
+
+```sh
+$ npm run build
 ```
 
 ## Testing
--- a/benchmark/parse.js
+++ b/benchmark/parse.js
@@ -9,7 +9,7 @@ var benchmarks = require('beautify-benchmark')
  * Globals for benchmark.js
  */
 
-global.cookie = require('..')
+global.cookie = require('../pkg')
 
 var suite = new benchmark.Suite()
 
--- a/src/index.js
+++ b/src/index.js
@@ -8,14 +8,6 @@
 'use strict';
 
 /**
- * Module exports.
- * @public
- */
-
-exports.parse = parse;
-exports.serialize = serialize;
-
-/**
  * Module variables.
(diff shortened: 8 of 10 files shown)


.gitignore
```diff
@@ -1,4 +1,4 @@
 coverage/
 node_modules/
 npm-debug.log
-package-lock.json
```
dougwilson: Please add this back and remove the `package-lock.json` file. It was ignored on purpose :)

pkg/README.md
```diff
@@ -0,0 +1,261 @@
+# cookie
```
dougwilson: Is there a method to implement this that does not require maintain two identical copies of the README in the source control? That seems like a lot to do in order to add this `export`/`import` feature.

pkg/dist-node/index.js
```diff
@@ -0,0 +1,189 @@
+'use strict';
```
dougwilson: Is there a method to implement this that does not require maintain multiple identical copies of the source code in the source control? That seems like a lot to do in order to add this `export`/`import` feature.

pkg/dist-src/index.js
```diff
@@ -0,0 +1,188 @@
+/*!
```
dougwilson: Is there a method to implement this that does not require maintain multiple copes of the source in the source control? That seems like a lot to do in order to add this `export`/`import` feature.

pkg/dist-web/index.js
```diff
@@ -0,0 +1,187 @@
+/*!
```
dougwilson: Is there a method to implement this that does not require maintain multiple copes of the source in the source control? That seems like a lot to do in order to add this `export`/`import` feature.

pkg/package.json
```diff
@@ -0,0 +1,51 @@
+{
```
dougwilson: Is there a method to implement this that does not require maintain multiple copies of the `package.json` in the source control? That seems like a lot to do in order to add this `export`/`import` feature.

jorgecasar:
WIP to solve build issues

package.json
```diff
@@ -13,27 +13,56 @@
...
     "node": ">= 0.6"
   },
   "scripts": {
     "bench": "node benchmark/index.js",
     "lint": "eslint --plugin markdown --ext js,md .",
-    "test": "mocha --reporter spec --bail --check-leaks test/",
-    "test-ci": "istanbul cover node_modules/mocha/bin/_mocha --report lcovonly -- --reporter spec --check-leaks test/",
-    "test-cov": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter dot --check-leaks test/"
+    "test": "mocha --recursive --require @babel/register --bail --check-leaks",
+    "test-ci": "cross-env NODE_ENV=test nyc npm test",
```
dougwilson: Can you please restore this back to `Istanbul`? Or at least can you explain why we need to change? Currently as far as I know `Istanbul` is working just fine, and it works on more versions of Node.js than `nyc` does, so helps us check coverage on our supported runtimes.

.babelrc
```diff
@@ -0,0 +1,17 @@
+{
+	"presets": [
```
dougwilson: Just needs spaces instead of tabs for indent in this file.

pkg/dist-node/index.js
jorgecasar: Yes, there are a tool that convert ES Module syntax (ESM) to Common.js and include type definitions for TypeScript.

pkg/dist-src/index.js
jorgecasar: Yes, there are a tool that convert ES Module syntax (ESM) to Common.js and include type definitions for TypeScript.

pkg/dist-web/index.js
jorgecasar: Yes, there are a tool that convert ES Module syntax (ESM) to Common.js and include type definitions for TypeScript.

pkg/package.json
jorgecasar: Yes, there are a tool that convert ES Module syntax (ESM) to Common.js and include type definitions for TypeScript.

.gitignore
jorgecasar: Ok, added again

pkg/README.md
jorgecasar: Yes, runing `npm run build` you build all pkg folder. I include it in the repo because I not sure how to build it in old versions of node.

package.json
jorgecasar: Changed, turned back istanbul.

.babelrc
jorgecasar: Ok, my editor betray me

jorgecasar:
With this change you have to publish on NPM the folder pkg, you can use `pack publish` to do it.

Do you want to include a script `"publish": "pack publish"` in `package.json` file?

pkg/package.json
dougwilson: Sorry, the question was how to not maintain multiple copies _in source control_.

For example if someone wants to contribute, how do they know which files yo modify?

dougwilson:
Thanks for the updates! If the only way to add this import/export stuff is to keep duplicate copies of things checked into source control, I'm going to respectfully decline adding the feature to this repo.

jorgecasar:
I will investigate how to build it with Travis and share the build with all jobs to test in all versions of node

dougwilson:
Thank you!!

wesleytodd:
I didn't realize that a PR had been opened for this already.  I am a big 👎 on this change.  Changing to ES modules add's nothing to this package for the maintainers or the users.  Why would we want to move forward with this?

wesleytodd:
I realize I am late to the party on this, sorry if I am creating noise where it is unnecessary.  As @dougwilson already said he is also (respectfully) 👎 on this.  I just wouldn't want @jorgecasar to waste (more?) time.  @jorgecasar is there something you think we have to gain by using ES module syntax in this package?  

I am very much of the opinion that no one should be using ES modules in anything other than experiments yet.  It is just not ready for prime time.  And converting existing working modules to it is not a good endeavor.  I would be happy to hear something contrary to this opinion :)

dougwilson:
It's not too late and always good to have another member opinion. I was out yesterday trying to enjoy some of the holiday weekend where I am, so didn't get to really respond too much to anything that occurred since then.

I only use require() myself and so don't truly know if import provides something extra. The original issue opened I really just wanted to know what it would take to support it, and a pull request was I guess the answer instead of words :) that is fine, of course.

The current state of this pull request does seem like a lot to do to add thus support, though.

jorgecasar:
@wesleytodd thanks for your feedback. ES Modules could be experimental in backend side, but it's very well supported in modern browsers (79.9% via [modules support in Can I use](https://caniuse.com/#feat=es6-module)). Web Components and most of the modern frameworks and libraries are using it.

IMHO, I guess that we have to push standards and migrate libraries to support them as soon as possible. Because if no body use the new standard we can discover corner cases or issues in the definition.

wesleytodd:
> ES Modules could be experimental in backend side

This is a backend module at its heart (and original intent).  So any decisions made for it's future would need to take that into account.  That being said, I see no reason we could not also support browsers, as nothing other than lack of test infrastructure is blocking it.

> IMHO, I guess that we have to push standards and migrate libraries to support them

I could not disagree more.  If a library has a reason other than "its new so better" to adopt a new standard, then great!  But with es modules it is actually *worse* than common js for many reasons.  Discover corner cases and issues in play modules, not important ones to the ecosystem like this one.

jorgecasar:
Since Bower was deprecated NPM become the Frontend package manager too. And we should try to get libs working in both environments. As I said in the issue https://github.com/jshttp/cookie/issues/87#issuecomment-464512109 There are frontend libraries using this library that doesn't work in module syntax and we have to use babel or webpack to make it work loosing the power of ES modules.

In the other hand, I close the PR because the discourse changed. We can move the discussion to #87 to avoid disinformation. There, I suggest to support ES modules by adding `module` attribute to the `package.json` with the path of the file using modules. There are ways to add this support without change the bases.

More info: https://www.pikapkg.com/about

anasouardini:
Importing it using `import cookie from 'cookie';` gives me `undefined`
Version: 1.0.1

wesleytodd:
@anasouardini this issue is old and closed. If you wish to open a new issue and explain your issue in more detail, please do.

PR #98, opened 2020-02-15
perf: put exports at the end of the file

### LINKED ISSUE #97: Export statement should be placed at the bottom of the script for editor support
Hello,
Intellij can not recognize the function of the cookie object, when the methods are not defined before the export statement. In this case these two lines should be flipped, so they are in the right order. First declaration, then assignment.

```js
// ...

exports.parse = parse;
exports.serialize = serialize;
```

tim-kilian:
Closes #97

## Changes
- First define methods and then assign it afterwards to exports

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,14 +8,6 @@
 'use strict';
 
 /**
- * Module exports.
- * @public
- */
-
-exports.parse = parse;
-exports.serialize = serialize;
-
-/**
  * Module variables.
  * @private
  */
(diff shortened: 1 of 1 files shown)


dougwilson:
Thanks, but I prefer them to be at the top so seeing them is obvious. It is 100% valid comon JS export usage and javascript. It sounds like perhaps a bug should be reported to your editor, unless you can provide official documentation from your editor that says this is required.

Issue #100, opened 2020-03-12
Support for sameSite === 'Unset'


AKushWarrior:
Various sites (ex. https://slader.com) don't set the sameSite cookie attribute, and instead leave it as 'Unset'. This currently tosses a TypeError, whereas behavior should be the same as 'none'.

dougwilson:
I'm not sure you are posting on the right repo. This module allow you to send either none or simply not set, neither throws a TypeError.

If you can provide the exact steps to reproduce the error, the version of this module you are using, and the exact error message, I can always reopen to investigate.

AKushWarrior:
Essentially, if you parse a cookie from a site (slader.com, as I mentioned above) and try to send a modified version, the parse function fails. It does so because of [this switch statement](https://github.com/jshttp/cookie/blob/aec1177c7da67e3b3273df96cf476824dbc9ae09/index.js#L163). I'm proposing adding 'Unset' as a case in the switch statement, so that trying to send the same cookie doesn't fail.

dougwilson:
I'm not sure I understand what you mean by "if you parse a cookie from a site", can you elaborate? This module only _parses_ the `"Cookie"` header (NOT the `"Set-Cookie"` header)...

AKushWarrior:
Sorry, let me clear some things up. I'm using the [serialize function](https://github.com/jshttp/cookie/blob/aec1177c7da67e3b3273df96cf476824dbc9ae09/index.js#L101) to serialize a set-cookie header. This header is read from a crawled site and has to be re-serialized with certain changed attributes. I'm not modifying the 'same-site' field, so I didn't expect this package to error when I attempted to re-serialize the header.

Using 'parse' was a total mental error by me. I've been mixing and matching on npm for a while now, and I got my terminology mixed up.

dougwilson:
So this module will error on _by design_ if the value you pass to `sameSite` option is not valid according to the specification of the `Set-Cookie` header. Having no `SameSite` attribute is valid, of course, and the API of this module is to not provide the `sameSite` option in that case (or use `sameSite: undefined`).

AKushWarrior:
https://tools.ietf.org/id/draft-ietf-httpbis-rfc6265bis-03.html#the-samesite-attribute Take a look at the last sentence of this. Values that are not listed should be treated as 'None', not an error, so as to maintain support for broken websites.

dougwilson:
[...]

PR #101, opened 2020-03-12
Update index.js to support "sameSite === 'Unset'"

AKushWarrior:
As per issue #100 ...

### DIFF
--- a/index.js
+++ b/index.js
@@ -173,6 +173,9 @@ function serialize(name, val, options) {
       case 'none':
         str += '; SameSite=None';
         break;
+      case 'Unset':
+        str += '; SameSite=None';
+        break;
       default:
         throw new TypeError('option sameSite is invalid');
     }


dougwilson:
This module only supports the values as defined in the RFC: https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#section-4.1.2.7

If you can link to the RFC that adds this new value, we can re-consider 👍

AKushWarrior:
Look at the last sentence of this section from IETF: https://tools.ietf.org/id/draft-ietf-httpbis-rfc6265bis-03.html#the-samesite-attribute

The default should not be a TypeError, it should be the same behavior as 'None'

dougwilson:
> The default should not be a TypeError, it should be the same behavior as 'None'

That wording is in regards to _parsing_ the attribute's value, not serializing.

AKushWarrior:
Okay, I see now. This PR and issue should be for my parsing library :/

Have a good day!

Issue #103, opened 2020-04-13
Max-Age are taking Infinity this is not valid for http specification


smondal:
Max-Age are showing "Infinity", If I give big number. This is not a valid value for Max-Age field.
https://tools.ietf.org/html/rfc6265#section-5.2.2

cookie.serialize('name', String("sandip"), {httpOnly: "sandip", maxAge: 60 ** 60 ** 24, expires: new Date()})
'name=sandip; Max-Age=**Infinity**; Expires=Mon, 13 Apr 2020 15:39:35 GMT; HttpOnly'

dougwilson:
Thanks, definitely a bug. It should throw on `Infinity` provided as `maxAge` option since there is no way to serialize that.

PR #104, opened 2020-04-14
Bug: invalid value Infinity for maxAge

smondal:
Regarding issue #103, We should handle +Infinity and -Infinity for maxAge value

### DIFF
--- a/index.js
+++ b/index.js
@@ -120,7 +120,11 @@ function serialize(name, val, options) {
 
   if (null != opt.maxAge) {
     var maxAge = opt.maxAge - 0;
-    if (isNaN(maxAge)) throw new Error('maxAge should be a Number');
+
+    if (isNaN(maxAge) || !isFinite(maxAge)) {
+      throw new TypeError('option maxAge is invalid')
+    }
+
     str += '; Max-Age=' + Math.floor(maxAge);
   }
 
--- a/test/serialize.js
+++ b/test/serialize.js
@@ -55,7 +55,13 @@ test('maxAge', function() {
     cookie.serialize('foo', 'bar', {
       maxAge: 'buzz'
     });
-  }, /maxAge should be a Number/);
+  }, /option maxAge is invalid/)
+
+  assert.throws(function () {
+    cookie.serialize('foo', 'bar', {
+      maxAge: Infinity
+    })
+  }, /option maxAge is invalid/)
 
   assert.equal('foo=bar; Max-Age=1000', cookie.serialize('foo', 'bar', {
     maxAge: 1000
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+unreleased
+==========
+
+  * Fix `maxAge` option to reject invalid values
+
 0.4.0 / 2019-05-15
 ==================
 


index.js
```diff
@@ -121,6 +121,7 @@ function serialize(name, val, options) {
   if (null != opt.maxAge) {
     var maxAge = opt.maxAge - 0;
     if (isNaN(maxAge)) throw new Error('maxAge should be a Number');
+    if (maxAge === Infinity || maxAge === -Infinity) throw new Error("Infinity is invalid value for maxAge")
```
dougwilson, line 124:
```suggestion
    if (!Number.isFinite(maxAge)) throw new Error('maxAge must be finite')
```

test/serialize.js
```diff
@@ -84,6 +84,18 @@ test('maxAge', function() {
...
+    cookie.serialize('foo', 'bar', {
+      maxAge: (Number.MAX_VALUE) * 2
+    });
+  }, /Infinity is invalid value for maxAge/);
+
+  assert.throws(function () {
+    cookie.serialize('foo', 'bar', {
+      maxAge: (-Number.MAX_VALUE) * 2
+    });
+  }, /Infinity is invalid value for maxAge/);
```
dougwilson, line 98:
```suggestion
  }, /-Infinity is invalid value for maxAge/)
```

test/serialize.js
```diff
@@ -84,6 +84,18 @@ test('maxAge', function() {
   assert.equal('foo=bar; Max-Age=3', cookie.serialize('foo', 'bar', {
     maxAge: 3.14
   }));
+
+  assert.throws(function () {
+    cookie.serialize('foo', 'bar', {
+      maxAge: (Number.MAX_VALUE) * 2
```
dougwilson, line 90:
```suggestion
      maxAge: Infinity
```

test/serialize.js
```diff
@@ -84,6 +84,18 @@ test('maxAge', function() {
...
+    });
+  }, /Infinity is invalid value for maxAge/);
+
+  assert.throws(function () {
+    cookie.serialize('foo', 'bar', {
+      maxAge: (-Number.MAX_VALUE) * 2
```
dougwilson, line 96:
```suggestion
      maxAge: -Infinity
```

dougwilson:
I made a few suggestion, let us know what you think.

index.js
jonchurch: I first thought of this as well, but MDN suggested that `Number.isFinite` isn't supported in IE.

I know this package is meant to be used serverside, and also I'm not sure at all if there are other things in here that IE doesn't support, but I figured there's likely someone running this in the frontend somewhere.

dougwilson: That seems like a fair thought, but (1) we don't support web browsers unless they are part of the module's CI and (2) this module seems almost useless in the browser context, as it is about reading cookies browsers _send_ and writing cookies browsers _receive_; it has no functionality for writing cookies browsers need to send or reading cookies browsers would receive.

dougwilson: Perhaps just use the `isFinite` global function?

dougwilson:
```suggestion
    if (!isFinite(maxAge)) throw new Error('maxAge must be finite')
```

jonchurch: That sounds like a good compromise 👍 

Here's at least one [package](https://www.npmjs.com/package/universal-cookie) that seems to be using this in the Browser 😨

dougwilson: Ultimately, this package is under the semver 0.y.z rules right now.

jonchurch: They're using it to [parse](https://github.com/reactivestack/cookies/blob/bb68224696a59d8089fa85aa4f116c205ff31a92/packages/universal-cookie/src/utils.ts#L22) and [serialize](https://github.com/reactivestack/cookies/blob/master/packages/universal-cookie/src/Cookies.ts#L79) cookies, apparently

dougwilson: Since this module is under the semver 0.y.z rules, there are no support rules; if they don't want to see it break, the best option here is that could contribute a web browser test suite to the module, otherwise we cannot properly make changes to this module without possibly breaking them. For all we know, they don't even support IE themselves, meaning if we broke IE, they wouldn't care.

It's one thing to make decisions in our module within the constraints we set on ourselves (Node.js versions supported, adherence to SemVer, etc.) but we can't really be upheld to unknown constraints (like whatever web browser environments are out there) since we don't know what folks even need on that front.

test/serialize.js
smondal: This is better solution

test/serialize.js
smondal: :+1:

smondal:
This is better solution. As per your suggestion fixes are added into pull request

smondal:
Travis is falling as isFinite is not supported by Nodejs 0.6 version

dougwilson:
Yea, I believe the updated recommendation was to use the `isFinite` global function instead of the `Number.isFinite` class function.

smondal:
Let me try with global isFinite function.

PR #119, opened 2021-01-13
Serialize object

export-mike:
I have added support for serializing objects to a cookie string. 

If there's added detail, tests etc that are needed, let me know. thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,16 @@ var setCookie = cookie.serialize('foo', 'bar');
 // foo=bar
 ```
 
+### cookie.serialize(cookieObject, options)
+
+Serialize a cookie object map into a `Set-Cookie` header string. the `object` argument is an object of key value pairs for each cookie, the options
+argument is an optional object containing additional serialization options.
+
+```js
+var setCookie = cookie.serialize({foo: 'bar'});
+// foo=bar
+```
+
 #### Options
 
 `cookie.serialize` accepts these properties in the options object.
--- a/index.js
+++ b/index.js
@@ -5,6 +5,8 @@
  * MIT Licensed
  */
 
+'use strict';
+
 /**
  * Module exports.
  * @public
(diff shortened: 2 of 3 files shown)


dougwilson (changes requested):
The output of this is not a valid header string. Please refer to the RFCs for what it should look like and browser testing. I was not able to make this work to set two cookies in a web browser. If this does work, perhaps the docs arw unclear. Can you provide an example app that demonstrates this functionality working I can test in a wrb browser?

export-mike:
I've updated to add the missing semi colon, to get this added at the correct time. Solution isn't ideal. It would require some further refactoring, so didn't want to deviate too much right now.

interesting though with regards to document.cookie vs Set-Cookie. Set-Cookie supports multiple. Whereas document.cookie does not.

I wasn't aware of this. how annoying! I guess by not supporting that we get expected behaviour in all environments. 

Not sure how we could proceed really - This is quite useful to have this in a server side context. 

Options

- Check for document.cookie - If it exists log an error, also add some details to the docs that this function will not work if you are passing it to document.cookie.

- create a serializeObject function - with a note in the docs saying it'll only work with Set-Cookie for example

- Do not make the change whatsoever.

export-mike:
I'm closing this PR as I don't have the capacity to continue further changes on this.

It feels like a useful change but only in a server context. document.cookie doesn't support multiple cookies in a single string

feel free to re-open I feel this is changing the libraries behaviour in a browser vs server context. its a trivial change to merge this in, if its useful - I don't think its useful after discovering the usage in a browser with document.cookie

spaceone:
I needed this change as well for a client library.
Cookies have to be split via `; ` (not only `;`) (RFC 6265 Section 4.2.1). This library doesn't contain functionality to support this but it would be achievable with that change.

`cookie.serialize({foo: "bar", bar: "baz"}) == 'foo=bar; bar=baz'`.
instead of `[cookie.serialize('foo', 'bar'), cookie.serialize('bar', 'baz')].join('; ')`

PR #120, opened 2021-02-03
fix a typo and some deprecation warning

mhmda-83:


dougwilson (changes requested):
It seems like this PR consists of at least two separate things; please split those out into separate PRs. For example, fixing the typo is straight forward and could have been merged immediately, but there are tests changes that cause CI to fail, so the PR cannot be merged due to that. If it is split, then the typo can be fixed while a resolution to the test changes are worked on.

Issue #127, opened 2021-04-19
ESM version


benmccann:
There was some previous discussion of this in https://github.com/jshttp/cookie/issues/87. However, I wanted to raise the issue since the JavaScript landscape has changed in the time since then and there are some considerations that were not raised in the original thread

Unbundled development is becoming very popular using tools like Snowpack, Vite, and WMR. These tools serve code in an unbundled fashion using purely ESM. Any library that is not distributed as ESM must be converted to ESM by these tools. That can be a somewhat error-prone process and in fact Vite currently chokes on this library (https://github.com/vitejs/vite/issues/2579). Those issues should be fixed in the development tools themselves, though we would avoid the issue if `cookie` were distributed in ESM. However, even if this issue is fixed, the lack of an ESM version means that development with these tools is slower because they must convert the library to ESM on first run of the project. It would be a better development experience to do the conversion just once before distributing the library rather than making every user do the conversion each time they run `npm install`

ESM is the standard and Node can now handle ESM natively. Node 8 and 10 support it via the `--experimental-modules` flag while Node 12 and newer support if out-of-the-box. At the end of this month, Node 10 will go EOL meaning every supported version of Node will support ESM out-of-the-box. It's quite possible at this point that it's no longer necessary to distribute a CJS version. However, if it's still preferred to support these older versions of node without requiring an extra CLI flag, it would be fairly straightforward to setup Rollup or some other process to distribute both a CJS and ESM version in the package.

Finally, supporting ESM would allow the library to be used in the browser and in isomorphic apps. This may not be something we wish to commit to supporting, but it could be a nice side-effect that it would be allowed

I'm happy to send a PR for this, but thought I'd check whether there's any openness to the idea first

dougwilson:
Hi @benmccann thanks for opening this issue! I don't see any reason why we cannot add ESM support, though there are a couple caveats here:

1. This module's goal is to support as many Node.js versions as possible, so there is no dropping almost any of the Node.js support in the foreseeable future, especially since this module is a part of Express.js and it needs to continue to support that.
2. Introducing a build step like rollup in this module to continue CJS is not just not going to be acceptable. I'm sure there must be some method to support CJS and ESM without that, as I believe I have seen other modules do this, so adding complication and the ability for security compromises in the build process is just not going to really be a good solution.

I hope that provides some constraints around the conversation. I think if we work together, someone who knows how to implement ESM can help add it to this module with the above constraints intact.

benmccann:
[...]

PR #128, opened 2021-04-19
Add ESM compatibility

benmccann:
Closes https://github.com/jshttp/cookie/issues/127


dougwilson (changes requested):
Thanks @benmccann ! I flagged this as `needs tests` since it seems there are no tests added. At the very least, I don't even know how to test by hand, but ideally the CI would perform the tests to prevent regressions.

dougwilson:
Also looks like `npm run lint` no longer lints the module's source, just the tests.

benmccann:
Thanks. I rebased against `master`. Hopefully that fixed the `lint` issue

I've added a test as well

benmccann:
I finally got the tests setup to test both the CJS and ESM versions, but it turns out this wrapper approach doesn't work with Vite

benmccann:
Although that seems to be a Vite bug. It's just ignoring that we're distributing an ESM version and trying to load the CJS version directly, which seems wrong. This PR still seems useful even if Vite is too broken to use it yet

PR #137, opened 2022-03-21
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/benchmark/parse-top.js
+++ b/benchmark/parse-top.js
@@ -43,5 +43,5 @@ function gencookies (num) {
     str += '; foo' + i + '=bar'
   }
 
-  return str.substr(2)
+  return str.slice(2)
 }
--- a/benchmark/parse.js
+++ b/benchmark/parse.js
@@ -70,5 +70,5 @@ function gencookies (num) {
     str += '; foo' + i + '=bar'
   }
 
-  return str.substr(2)
+  return str.slice(2)
 }


dougwilson:
Thanks @CommanderRoot ! Looks like I added more since the last time I removed them in 04be428b438605b48ad6af503227b817c07b9b52 . Can you add a eslint rule that will keep them from coming back in the future, please?

CommanderRoot:
I did some [digging ](https://stackoverflow.com/questions/40977713/how-to-forbid-a-specific-named-function-with-eslint)and this seems to be the easiest way to add an eslint rule for disallowing substr.
There are some other ways as well but I'm no eslint expert

dougwilson:
Ah, thanks! I wonder if this rules could simply be added directly to StandardJS (https://standardjs.com/), if you are up for it? It would be a huge help, at least though the dozen or so projects you just PR'd tonight on me :) It already detects Node.js deprecated API usage.

CommanderRoot:
I wasn't aware of StandardJS until now. I have looked into it a bit and found a discussion https://github.com/standard/standard/issues/922 and a PR https://github.com/standard/eslint-config-standard/pull/185 which partly integrates https://github.com/sindresorhus/eslint-plugin-unicorn which has a "slice > substr or substring" rule.
The PR however is open for nearly a year so I have not much hope there.
In general I like the idea though to prevent further `substr()` usage in new code.

dougwilson (approved):

PR #143, opened 2022-07-14
perf(parse): remove additional slice for quoted values

kurtextrem:
see https://github.com/fastify/fastify-cookie/pull/194#issuecomment-1184444263

Before & after as diff:
https://www.diffchecker.com/hO2It0SH on node v16 (`unquote` -> +300 000 op/s)

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,15 +54,17 @@ function parse(str, options) {
   var dec = opt.decode || decode;
 
   var index = 0
+  var eqIdx = 0
+  var endIdx = 0
   while (index < str.length) {
-    var eqIdx = str.indexOf('=', index)
+    eqIdx = str.indexOf('=', index)
 
     // no more cookie pairs
     if (eqIdx === -1) {
       break
     }
 
-    var endIdx = str.indexOf(';', index)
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -72,16 +74,13 @@ function parse(str, options) {
...
     // only assign once
     if (undefined === obj[key]) {
-      var val = str.slice(eqIdx + 1, endIdx).trim()
-
-      // quoted values
-      if (val.charCodeAt(0) === 0x22) {
-        val = val.slice(1, -1)
-      }
+      var val = (str.charCodeAt(eqIdx) === 0x22)
+        ? str.slice(eqIdx + 1, endIdx - 1).trim()
```
blakeembrey, line 82: I like this change for performance, but I think this changes the behavior. Before it trimmed whitespace around potential quotes, but now it'd trim whitespace within quotes.

blakeembrey:
Going to close this out, I managed to get an extra 10% perf boost across the board by instead iterating using a loop over whitespace (no more `trim` at all) in https://github.com/jshttp/cookie/pull/170, and it maintains backward compatibility with the quote characters.

PR #145, opened 2022-10-02
Fix handle special Object properties

repsac-by:
Handling of special `Object` properties such as `toString`, `__proto__`, etc..

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,7 +49,7 @@ function parse(str, options) {
     throw new TypeError('argument str must be a string');
   }
 
-  var obj = {}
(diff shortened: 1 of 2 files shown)


dougwilson:
Hello, and thank you for your pull request. Unfortunately this is not going to be able to land any time soon, as downstream expects the returned object to inherit from `Object.prototype`, which is why cookies named things like `__proto__` are ignored by this module. Returning an object that includes these named-properties ends up causing issues with consumers due to various merge utilities. Ignoring these named-cookies was done to resolve a security report.

PR #151, opened 2023-05-14
added support for partitioned key as per new Chrome regulations

punithk:
As per new Chrome regulations for third party cookie, which would be blocked, partitioned must be sent to ensure a separate cookie jar in maintained.

Refer new changes for third party cookie here: <https://developer.chrome.com/en/docs/privacy-sandbox/chips/>

PR #153, opened 2023-08-30
Add support for Partitioned.

pazguille:
@dougwilson I sent a complete PR for #151 🚀

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,8 @@
+unreleased
+==========
+
+  * Add `partitioned` option
+
 0.5.0 / 2022-04-11
 ==================
 
--- a/README.md
+++ b/README.md
@@ -107,6 +107,17 @@ The given number will be converted to an integer by rounding down. By default, n
 `maxAge` are set, then `maxAge` takes precedence, but it is possible not all clients by obey this,
 so if both are set, they should point to the same date and time.
 
+##### partitioned
+
+Specifies the `boolean` value for the [`Partitioned` `Set-Cookie`](rfc-cutler-httpbis-partitioned-cookies)
+attribute. When truthy, the `Partitioned` attribute is set, otherwise it is not. By default, the
+`Partitioned` attribute is not set.
+
+**note** This is an attribute that has not yet been fully standardized, and may change in the future.
+This also means many clients may ignore this attribute until they understand it.
+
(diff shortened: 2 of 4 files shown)


pazguille:
Hi @dougwilson Any news?

dougwilson:
Oh, didn't realize there was another PR for this; I checked the original a few days ago. I will take a look at this one ASAP

pazguille:
Hi @dougwilson, Any news? We already need this feature.

vladimiratanasov:
@pazguille thank you for opening this PR. May I suggest adding the following to the readme as well, it is important for setting the Partitioned state

```
Partitioned cookies must be set with Secure and Path=/. In addition, it is recommended to use the __Host prefix when setting partitioned cookies to make them bound to the hostname and not the registrable domain.
```
Source: https://developer.mozilla.org/en-US/docs/Web/Privacy/Partitioned_cookies#how_does_chips_work

dougwilson:
We don't do that with others flags like `samesite`, though. Do we feel that is a necessary thing to do? Probably should decide that before we merge bc it will be hard to add that in later without major version bumps and stuff. The samesite stuff kept changing on us (it didn't even start out with that name) as well, so I wouldn't be surprised if partitioned changes over time too, so may be better to just leave it as this PR is at least until the draft gets ironed out I think, right?

vladimiratanasov:
@dougwilson, makes sense. A disclaimer in the readme might be enough, stating that this library does not validate fields that partitioned depends on and this should be done on the client side. Once all browsers adapt a common standard (if ever), a validation can be added.

pazguille:
Is this necessary for merge this PR? Please let me know. We need this to update our cookies.

vladimiratanasov:
@dougwilson are you ok with merging this PR and releasing it? We will then have to wait for express to adopt the change, at least in our case since we don't use the package directly, and that might take a while.

nicob-29:
@dougwilson, 
Any Update ?
This PR is blocked other products .... :-(

dougwilson:
Sorry I was looking into getting the CI passing on the PR and life got in the way. I think I got it figured out though and should have it releaaed soon. Cannot merge a PR where the CI is failing.

dougwilson:
They are all fixed now. I just need to rebasw and merge this when I get home, just ran out of time yesterday.

dougwilson (approved):

Issue #152, opened 2023-06-22
Native ESM Support (no new build dependencies required)


PaulKiddle:
Hi thanks for this great project.

I know this has been discussed a couple of times before but not resolved. I'm using Rollup for my project and `cookie` is currently the only dependency that doesn't have ESM support, meaning I have to install the CJS plugin just for this package. It's fine, but ideally I'd like to simplify my build process even more by removing it.

I'm aware one of the reasons for not adding ESM support is that you don't want to add any build dependencies, which is fair enough. But there is a very simple way to add ESM support to this package just by concatenating extra lines at the start and end, which would require no build steps.

Since the only CJS-specific feature you're using is assigning properties to the `exports` object, basic ESM support would require creating a file like this:

```js
const exports = {};

/* ...CJS code goes here... */

export default exports;
```

This is easy enough to create using a common shell command like `cat`.

dougwilson:
Thank you, that is interesting. So you're saying I can just add those lines to the top and bottom of the `index.js` file and that is enough?

PaulKiddle:
That's correct - consumers would then be able to import the library using `import cookie from 'cookie'`;

If you want to support importing the individual functions too you'd need to explicitly export those:

```javascript
export const parse = exports.parse;
export const serialize = exports.serialize;
```

The consumer would then be able to use `import { parse, serialize } from 'cookie';`.

I believe there's also a config you'd need to add to the `package.json` that points to this esm version of the library.

Sammy-T:
@PaulKiddle You seem to already have a good handle on how this could be implemented. Would you be interested in creating a PR?

PaulKiddle:
Yep, if you're happy to go down this route I'll set up a PR when I get some free time

jorgecasar:
Thanks to accept the migration 4 years later I suggest it (#87)

blakeembrey:
Closing this out as the package is going to stay CJS only for now.

pi0:
We have been maintaining [unjs/cookie-es](https://github.com/unjs/cookie-es) for a few years now.

---

Just to be clear I am super thankful to the authors and maintainers of the amazing `cookie` package, this fork means no harm to the ecosystem, we try to keep the package in sync and backport any changes if they happen upstream.

kettanaito:
Hi! Can we please reconsider this? 

[...]

PR #154, opened 2023-10-17
Add ESM build script

PaulKiddle:
Adds a script to create the ESM version of the code, fixing #152 

The script is automatically run on `prepare` (i.e. when packaging for release or installing directly from git) on npm versions ≥ 4
The generated file is not committed to the repository

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ coverage/
 node_modules/
 npm-debug.log
 package-lock.json
+index.mjs
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ coverage/
 node_modules/
 npm-debug.log
 package-lock.json
-index.mjs
--- a/index.mjs
+++ b/index.mjs
@@ -1,0 +1,277 @@
+const exports = {}
+export {
+  parse,
+  serialize,
+  exports as default
+}
+
+/*!
+ * cookie
+ * Copyright(c) 2012-2014 Roman Shtylman
+ * Copyright(c) 2015 Douglas Christopher Wilson
+ * MIT Licensed
+ */
+
+'use strict';
+
+/**
+ * Module exports.
+ * @public
+ */
+
+exports.parse = parse;
+exports.serialize = serialize;
+
+/**
+ * Module variables.
+ * @private
+ */
+
+var __toString = Object.prototype.toString
+
+/**
+ * RegExp to match field-content in RFC 7230 sec 3.2
+ *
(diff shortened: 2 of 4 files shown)


dougwilson:
Awesome, thank you! I know there has been past discussions, so not sure if you have seen all of it, but a couple things for this PR i noticed missing, if you can think of how we can solve:

1. I added needs tests bc we want to make sure that however we are generating this actually reaults in an importable module in our ci insteqd of waiting for users to open an issue after a publish. Also may need tweaks down the line and so having a safety net in tests is always desirable.
2. Many of the security vendors that stare at the Express.js source code explicitly want us to have the files in the tarball on npm to be the exact files checked into the github repository at the commit hash that is in the npm metadata, so all these express used module we try to uphold that as best as we can, to keep their score high (as typically my understanding from them is untracked code in npm tarballs is considered suspicious by nature).

In addition, one of the pervious issues was that why does this build step need to exist at all? Node.js provides a way to have both cjs ans esm without any build steps needed. The last time it was due to Vite but the user decided it was a Vite bug and not to fix it here. Does that apply still, or can we simply follow the Node.js docs like https://nodejs.org/api/packages.html#approach-1-use-an-es-module-wrapper ?

Myself and I'm sure anyone else consuming esm is really thankful.

PaulKiddle:
Thanks for the feedback, those changes make sense.

As for why this is needed, my use case was Rollup which by default doesn't support common js (even via the wrapper method suggested in the Node docs). There is a plugin to enable it which I currently use, but if I can reduce my dependencies and config complexities, I would rather do that.

Plus I believe that given ESM modules have been the standard for a while we will start to see more build tools that do not support CJS, so being prepared for that possibility would be beneficial.

PaulKiddle:
@dougwilson It should be simple to run all the tests twice (once for each export type) by wrapping them in a function or loop (any preference?)
Would probably also need to distinguish if a test is running in ESM or CJS mode by adding something to the `describe` and/or `it` messages. I think just adding litterally `(esm)` or `(cjs)` to the `describe`s should suffice - does that sound fine?

PaulKiddle:
FYI I am waiting for confirmation that this approach sounds good before I go ahead and do the work to implement this

blakeembrey:
Hi @PaulKiddle, thanks for the PR! I'm trying to tidy up some older PRs in this repo and I'm going to close this one as we won't be adding dual package publishing in the immediate future. As far as I understand, most bundlers support CommonJS. I think this will make sense to revisit if/when we go ESM only.

PR #163, opened 2024-05-03
Feature: Add standard codes to serialize TypeErrors

MaoShizhong:
As requested in #162 as a non-breaking change prior to a change of error messages for a future major release, standard Node error codes have been added to each of the `serialize` function's TypeError throws.

This should allow for a period of migrating for library users to error-handling via the `.code` error property, as opposed to the sole `.message` property, before they change (potentially breaking).

## This PR

- Amends existing `serialize` tests to test for the correct error message and error code.
- Adds a sameSite test for non-string/true argument values.
- Adds `.code` properties with standard Node error codes to each error throw in the `serialize` function.

### DIFF
--- a/index.js
+++ b/index.js
@@ -113,17 +113,23 @@ function serialize(name, val, options) {
   var enc = opt.encode || encode;
 
   if (typeof enc !== 'function') {
-    throw new TypeError('option encode is invalid');
+    var encodeError = new TypeError('option encode is invalid')
+    encodeError.code = 'ERR_INVALID_ARG_TYPE'
+    throw encodeError
   }
 
   if (!fieldContentRegExp.test(name)) {
-    throw new TypeError('argument name is invalid');
+    var nameError = new TypeError('argument name is invalid')
+    nameError.code = 'ERR_INVALID_ARG_VALUE'
+    throw nameError;
   }
 
   var value = enc(val);
 
   if (value && !fieldContentRegExp.test(value)) {
-    throw new TypeError('argument val is invalid');
+    var returnError = new TypeError('argument val is invalid')
+    returnError.code = 'ERR_INVALID_RETURN_VALUE'
+    throw returnError;
   }
 
   var str = name + '=' + value;
@@ -132,7 +138,9 @@ function serialize(name, val, options) {
     var maxAge = opt.maxAge - 0;
 
     if (isNaN(maxAge) || !isFinite(maxAge)) {
-      throw new TypeError('option maxAge is invalid')
+      var maxAgeError = new TypeError('option maxAge is invalid')
+      maxAgeError.code = 'ERR_INVALID_ARG_VALUE'
+      throw maxAgeError;
     }
 
     str += '; Max-Age=' + Math.floor(maxAge);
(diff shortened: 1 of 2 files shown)


MaoShizhong:
Looks like asserting throws with a POJO was breaking backwards compatibility with the action runs. This change should hopefully be backwards compatible with the versions now.

wesleytodd (commented):
These CI issues are not related. In general this is *great*, but I think we should probably clean up the inconsistencies in formatting and also there are a few nits for after we get the CI passing.

MaoShizhong:
Could I get some guidance on exactly what needs fixing here for the action runs? I'm a little confused as to why some tests that passed last run now suddenly fail after only changing the indentation for one of the files (which understandably was causing some of the later Node version tests to fail)
As far as I'm concerned, the original issue regarding the test object/callback has been resolved, as has the indentation and on my end, I remember Node v0.10 and v0.12 passing respectively.

wesleytodd:
> Could I get some guidance on exactly what needs fixing here for the action runs?

Ah, sorry yeah I should have been more clear. I don't think you need to fix this in this PR. I *think* we have been considering just dropping the coveralls stuff, but it is unclear what we want bigger picture. I don't want to block landing things on this kind of unrelated stuff, so what I was thinking is I would run these tests locally once to check, but I dont see anything which *should* break so I hope we could just skip the check and *force* re-run each test run since it got past that part.

index.js
```diff
@@ -132,33 +140,44 @@ function serialize(name, val, options) {
...
 
     str += '; Path=' + opt.path;
   }
 
   if (opt.expires) {
     var expires = opt.expires
-
-    if (!isDate(expires) || isNaN(expires.valueOf())) {
-      throw new TypeError('option expires is invalid');
+    var expiresError = new TypeError('option expires is invalid')
```
jonchurch, line 173: allocating an error object when it's not needed incurs a very small performance hit, due to the stack trace generation

This is one of those instances where following Dont Repeat Yourself isn't worth the perf cost. We save a tiny amount of repetition, but at a performance cost.

MaoShizhong:
Just want to clarify what the desired way forward would be.
I can update this PR with the error codes e.g.
```ts
// src/index.ts:250
  if (!cookieNameRegExp.test(name)) {
    const nameError = new TypeError(`argument name is invalid: ${name}`);
    nameError.code = 'ERR_INVALID_ARG_VALUE';
    throw nameError;
  }
```
Though since we're now using TS, I'd need to handle `.code` not being a property on the `TypeError` interface. How would you like me to proceed with that?

Also perfectly happy to have this PR superceded by #208, that's not a problem by any means.

florafinessesbeauty:
It looks like you're dealing with a complex pull request related to the cookie package and ensuring TypeErrors include standard Node error codes. This kind of meticulous work is essential for robust error handling and maintaining backward compatibility.

To summarize the key points:

Error Codes in TypeErrors: The pull request adds .code properties to TypeError throws in the serialize and parse functions. This allows for more granular error handling.

Testing: Adjustments to existing tests ensure that they check for both the correct error message and the new error codes.

Performance Considerations: There's an ongoing discussion about the performance implications of creating error objects when not strictly necessary, which might impact performance in deep stacks.

Conflicting Tests: Some continuous integration (CI) tests are failing or have been canceled, particularly on older Node.jsversions. This needs further investigation to determine the exact cause and ensure that the changes do not break existing functionality.

test/compare-error.js
```diff
@@ -1,13 +1,13 @@
...
+    ) {
+      return true;
     }
-  }
+  };
+}
 
-  module.exports = compareError;
\ No newline at end of file
+module.exports = compareError;
```
vhf2jghn76-sys, line 13: https://weplay.run/A/qe4SfM

PR #166, opened 2024-07-16
Fix exports for some bundlers

proudparrot2:
I've been using this package in a javascript library with esbuild and it's worked fine, but for some reason, rspack can't resolve it. Adding `"main": "index.js"` to package.json fixed this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "SECURITY.md",
     "index.js"
   ],
+  "main": "index.js",
   "engines": {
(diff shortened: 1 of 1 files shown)


wesleytodd:
While this is a good change and likely to land here, you might want to report that as a bug to them. The default behavior when `main` is missing is to use `index.js`: https://nodejs.org/docs/latest/api/modules.html#folders-as-modules

PR #167, opened 2024-07-17
fix: narrow the validation of cookies to match RFC6265

### LINKED ISSUE #165: incorrect validation regexes in serialize
The code here:
https://github.com/jshttp/cookie/blob/master/index.js#L119-L125

... tests cookie names and values against the field-content regex from RFC7230. Unfortunately, that's the wrong pattern; that's the regex for the whole header value, not the individual parts. The cookie name and value per https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1 use `token` from rfc7230 for the name, and a specific
pattern in rfc6265 for the value. While the default encode function will keep values within the required pattern, it can be overridden, so the validation can fail on user-supplied encoders; the name validation is always wrong.

It's pretty easy to demonstrate:

```
cookie = require('cookie');

function identity(s) {
  return s;
}

tests = {
  goodname: 'goodvalue',
  'bad name': 'bad value',
  'bad=name': 'good=value',
  'bad;name': 'bad;value',
  'bad☹name': 'bad;value',
  name1: 'bad☹value',
  name2: 'bad"value',
  name3: '"goodvalue"',
};
// https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1 refers
// to https://datatracker.ietf.org/doc/html/rfc2616#section-2.2
// which is obsoleted and replaced by the token definition in
// https://www.rfc-editor.org/rfc/rfc7230#appendix-B
tokenRegex = /^[!#$%&'*+.^_`|~0-9A-Za-z-]+$/;
// https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1
cookieValueRegex =
  /^("?)[\u0021\u0023-\u002B\u002D-\u003A\u003C-\u005B\u005D-\u007E]*\1$/;

for (const [name, value] of Object.entries(tests)) {
  var validName = tokenRegex.test(name);
  var validValue = cookieValueRegex.test(value);
  var serialized;
  try {
    serialized = cookie.serialize(name, value, { encode: identity });
  } catch (error) {
    serialized = error;
  }
  console.log(
    `name: ${name} (${validName}) value: ${value} (${validValue}) serialized: ${serialized}`
  );
}
```

(demonstrating both good/bad values where the true regexes and cookie agree, and a handful of the many values where cookie would incorrectly say it's valid)

Discovered looking for whatever was leaking whitespace into our cookies; I'm not certain cookie did it, but seeing that the validation allowed whitespace jumped out at me.

bewinsnw:
fixes #165. Previously, validation on serialization used the field-content regexp from RFC7230 to perform all checks. However, that is the regexp for the cookie header as a whole, but each individual part of a cookie has a tighter restriction.

In the bug report #165 I demonstrated that whitespace in names and values was invalid but allowed by the field-content pattern. In this code I started by adding tests for those two cases, then added the expressions derived from the RFC; the relevant portions of the RFC have been inlined as comments.

Removing the field-content regexp unearthed that as well as those two it was being used to validate domain names and paths. Paths are fairly unrestricted (being any ascii character except control characters and semicolon) but the code was wrong here too - it allowed all 8-bit characters not just 7-bit characters.

Domain names have a well recognised pattern, there are only a couple of gotchas: RFC6265 explicitly says RFC1123 applies, so the leading character of a label is allowed to be a digit; and while I wondered for a moment if this should allow things like [::1] the domain matching section of https://datatracker.ietf.org/doc/html/rfc6265#section-5.1.3 teaches that domain values are _not_ ip addresses.

### DIFF
--- a/index.js
+++ b/index.js
@@ -65,6 +65,7 @@ var cookieValueRegExp = /^("?)[\u0021\u0023-\u002B\u002D-\u003A\u003C-\u005B\u00
  *                     upper case and a through z in lower case
  * <digit>           = any one of the ten digits 0 through 9
  */
+ 
 var domainValueRegExp = /^([a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?)([.][a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?)*$/i;
 
 /**
--- a/index.js
+++ b/index.js
@@ -26,7 +26,7 @@ var __toString = Object.prototype.toString
  * RegExp to match cookie-name in RFC 6265 sec 4.1.1
  * This refers out to the obsoleted definition of token in RFC 2616 sec 2.2
  * which has been replaced by the token definition in RFC 7230 appendix B.
- * 
+ *
  * cookie-name       = token
  * token             = 1*tchar
  * tchar             = "!" / "#" / "$" / "%" / "&" / "'" /
@@ -36,7 +36,7 @@ var __toString = Object.prototype.toString
 
 var cookieNameRegExp = /^[!#$%&'*+.^_`|~0-9A-Za-z-]+$/;
 
-/** 
+/**
  * RegExp to match cookie-value in RFC 6265 sec 4.1.1
  *
  * cookie-value      = *cookie-octet / ( DQUOTE *cookie-octet DQUOTE )
@@ -46,7 +46,7 @@ var cookieNameRegExp = /^[!#$%&'*+.^_`|~0-9A-Za-z-]+$/;
  *                     ; and backslash
  */
 
-var  cookieValueRegExp = /^("?)[\u0021\u0023-\u002B\u002D-\u003A\u003C-\u005B\u005D-\u007E]*\1$/;
+var cookieValueRegExp = /^("?)[\u0021\u0023-\u002B\u002D-\u003A\u003C-\u005B\u005D-\u007E]*\1$/;
 
 /**
  * RegExp to match domain-value in RFC 6265 sec 4.1.1
@@ -65,7 +65,7 @@ var  cookieValueRegExp = /^("?)[\u0021\u0023-\u002B\u002D-\u003A\u003C-\u005B\u0
  *                     upper case and a through z in lower case
  * <digit>           = any one of the ten digits 0 through 9
  */
-var  domainValueRegExp = /^([a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)([.][a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/;
+var domainValueRegExp = /^([a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?)([.][a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?)*$/i;
 
 /**
  * RegExp to match path-value in RFC 6265 sec 4.1.1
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -23,14 +23,57 @@ exports.serialize = serialize;
...
- * obs-text      = %x80-FF
+ * cookie-name       = token
+ * token             = 1*tchar
+ * tchar             = "!" / "#" / "$" / "%" / "&" / "'" /
+ *                     "*" / "+" / "-" / "." / "^" / "_" /
+ *                     "`" / "|" / "~" / DIGIT / ALPHA
  */
 
-var fieldContentRegExp = /^[\u0009\u0020-\u007e\u0080-\u00ff]+$/;
+var cookieNameRegExp = /^[!#$%&'*+.^_`|~0-9A-Za-z-]+$/;
```
wesleytodd, line 37: This is very much a nitpick, sorry, but could we maintain the ordering from the spec section in the regex and use escaping instead of the "end" trick? I just find that trick always throws me off when trying to visually parse groups like this which also include ranges.

```suggestion
var cookieNameRegExp = /^[!#$%&'*+.^_\-`|~0-9A-Za-z]+$/;
```

wesleytodd (changes requested):

wesleytodd:
Thanks, this VERY well structured from the comments to the new regular expressions, great work. CI is passing, but I want to wait a bit and see if anyone else on the project as opinions on this. My one review comment is not intended to block, so if you feel strongly about that form of regex please tell me.

index.js
bewinsnw: I've no objections to fixing it that way, makes sense; but NB I didn't commit your suggestion because that's not the order in the spec either!

wesleytodd (approved):

IamLizu (commented):
LGTM :+1:

index.js
```diff
@@ -23,14 +23,57 @@ exports.serialize = serialize;
...
+ * <letter>          = any one of the 52 alphabetic characters A through Z in
+ *                     upper case and a through z in lower case
+ * <digit>           = any one of the ten digits 0 through 9
+ */
+var domainValueRegExp = /^([a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?)([.][a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?)*$/i;
+
+/**
+ * RegExp to match path-value in RFC 6265 sec 4.1.1
+ *
+ * path-value        = <any CHAR except CTLs or ";">
```
bewinsnw, line 73:
```suggestion
 * path-value        = <any CHAR except CTLs or ";">
 * CHAR                =  %x01-7F
 *                              ; defined in RFC 5234 appendix B.1
```

I realised reading it back that this definition of CHAR (7bit not 8bit) might not be obvious.

index.js
```diff
@@ -23,14 +23,59 @@ exports.serialize = serialize;
...
+ * <label>           = <let-dig> [ [ <ldh-str> ] <let-dig> ]
+ *                     Labels must be 63 characters or less.
+ *                     'let-dig' not 'letter' in the first char, per RFC1123
+ * <ldh-str>         = <let-dig-hyp> | <let-dig-hyp> <ldh-str>
+ * <let-dig-hyp>     = <let-dig> | "-"
+ * <let-dig>         = <letter> | <digit>
+ * <letter>          = any one of the 52 alphabetic characters A through Z in
+ *                     upper case and a through z in lower case
+ * <digit>           = any one of the ten digits 0 through 9
+ */
```
bewinsnw, line 67:
```suggestion
 */
 
```
missing whitespace after the comment is inconsistent with the rest  of these blocks. Since I've made another edit anyway, fixing this

wesleytodd:
Hey, can we fix up the lint issues? After that I think we can land this and update in both `express` v4 and v5

bewinsnw:
> Hey, can we fix up the lint issues? After that I think we can land this and update in both `express` v4 and v5

🤦 that's what I get for doing those last two commits in the online editor. Fixed, thanks

bewinsnw:
I think that coveralls just glitched out. I can see at https://coveralls.io/builds/68891763 that the coveralls run was successful, but it looks like it never reported back to github. Just needs a kick.

PR #169, opened 2024-09-25
fix(serialize): added regex validation for cookie name

hdtmccallie:
A vulnerability exists for versions < 0.2.1 as the cookie name was not validated nor encoded to avoid 'separator' characters which lead to the ability for cookie names to be exploited for arbitrary cookie injection by passing in delimiters in the cookie name string. Starting with v0.2.1 cookie names are validated the same as cookie values per [RFC 7230 sec 3.2.6][1]

This fix adds a regex test based on [RFC 6265 sec 4.1.1][2] and [RFC 2616 sec 2.2][3] to check for valid characters allowed in cookie names. It throws an error if the criteria are not met as opposed to sanitizing the value which could lead to silent bugs and the potential for cookie name collision as different name strings could be santized to the same resultant value.

Tests also provided to ensure valid cookie names work, invalid ones error as well as a couple of attack strings.

[1]: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6
[2]: https://datatracker.ietf.org/doc/html/rfc2616#section-2.2
[3]: https://datatracker.ietf.org/doc/html/rfc2616#section-2.2

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,6 +21,36 @@ var decode = decodeURIComponent;
 var encode = encodeURIComponent;
 
 /**
+ * RegExp to match cookie-name in RFC 6265 sec 4.1.1
+ *
+ * cookie-name = token (as defined in RFC 2616 sec 2.2)
+ *
+ * token = 1*<any CHAR except CTLs or separators>
+ *
+ * Character classes:
+ *  \x21: !
+ *  \x23-\x27: # to ' (excluding " which is \x22).
+ *  \x2A-\x2B: * and +.
+ *  \x2D-\x2E: - and ..
+ *  \x30-\x39: 0 to 9 (digits).
+ *  \x41-\x5A: A to Z (uppercase letters).
+ *  \x5E-\x7A: ^ to z (excluding backslash and a few separators).
+ *  \x7C: |
+ *  \x7E: ~.
+ *
+ * separators     = "(" | ")" | "<" | ">" | "@"
+ *                | "," | ";" | ":" | "\" | <">
+ *                | "/" | "[" | "]" | "?" | "="
+ *                | "{" | "}" | SP | HT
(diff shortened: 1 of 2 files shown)


pi0:
dear @hdtmccallie do you mind to also port this for [cookie-es](https://github.com/unjs/cookie-es)? 🙏🏼

hdtmccallie:
> dear @hdtmccallie do you mind to also port this for [cookie-es](https://github.com/unjs/cookie-es)? 🙏🏼

@pi0 - it appears that cookie-es already [has this protection](https://github.com/unjs/cookie-es/blob/main/src/cookie/serialize.ts#L39).

I forgot to mention that this only applies to versions of cookie < v0.2.1.

blakeembrey:
@hdtmccallie Since this is an issue prior to 0.3, I'm assuming the correct path to upgrade for users would be to upgrade to 0.3? Is there something preventing that in a way that makes you want to patch 0.2? Just worried I'm missing something.

hdtmccallie:
> @hdtmccallie Since this is an issue prior to 0.3, I'm assuming the correct path to upgrade for users would be to upgrade to 0.3? Is there something preventing that in a way that makes you want to patch 0.2? Just worried I'm missing something.

@blakeembrey - This is meant to patch 0.1.3 in support of Express 3.21.2. I was operating under the assumption that an upgrade from 0.1.x to 0.2 or 0.3 could include breaking changes since the library hasn't reached 1.0 status yet. If this isn't the case and migrations between minor versions are safe then this fix might be unnecessary. Can you confirm? Multiple versions of Express 3, 4, 5 have dependencies of cookie 0.1.x.

hdtmccallie:
Closing this as there is more to the story and we need to regroup and approach this differently.

PR #171, opened 2024-10-02
test(serialize): additional tests for name, domain and path RFC validations

hdtmccallie:
These tests better align with the proper RFC rules for the cookie attributes of `name`, `domain`, and `path`.

These test are meant to be implemented along with [PR #167][1] that adds more fine-grained validations based on RFC rules.

[1]: https://github.com/jshttp/cookie/pull/167

### DIFF
--- a/test/serialize.js
+++ b/test/serialize.js
@@ -90,7 +90,7 @@ describe('cookie.serialize(name, value, options)', function () {
       });
     });
 
-    it('should throw for invalid value', function () {
+    it('should throw for invalid domain', function () {
       const invalidDomains = [
         'example.com\n',
         'sub.example.com\u0000',
@@ -206,7 +206,6 @@ describe('cookie.serialize(name, value, options)', function () {
   })
 
   describe('with "path" option', function () {
-
     it('should serialize path', function () {
       const validPaths = [
         '/',
@@ -248,8 +247,7 @@ describe('cookie.serialize(name, value, options)', function () {
         );
       });
     });
-
-  })
+  });
 
   describe('with "priority" option', function () {
     it('should throw on invalid priority', function () {
--- a/test/serialize.js
+++ b/test/serialize.js
@@ -17,23 +17,98 @@ describe('cookie.serialize(name, value)', function () {
     assert.equal(cookie.serialize('foo', ''), 'foo=')
   })
 
+  it('should serialize valid name', function () {
+    const validNames = [
+      'foo',
+      'foo!bar',
+      'foo#bar',
+      'foo$bar',
+      `foo'bar`,
+      'foo*bar',
+      'foo+bar',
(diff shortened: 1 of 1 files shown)


test/serialize.js
```diff
@@ -17,24 +17,98 @@ describe('cookie.serialize(name, value)', function () {
...
+      'foo\n',
+      'foo\u280a',
+      'foo/foo',
+      'foo,foo',
+      'foo;foo',
+      'foo@foo',
+      'foo[foo]',
+      'foo?foo',
+      'foo:foo',
+      'foo!foo',
```
blakeembrey, line 55: This was in the valid name list.

test/serialize.js
```diff
@@ -17,24 +17,98 @@ describe('cookie.serialize(name, value)', function () {
...
+      'foo.bar',
+      'foo^bar',
+      'foo_bar',
+      'foo`bar',
+      'foo|bar',
+      'foo~bar',
+      'foo7bar',
+    ];
+
+    validNames.forEach((name) => {
```
blakeembrey, line 39: This project runs all the way back to node 0.x so needed to avoid the new syntax.

test/serialize.js
```diff
@@ -133,14 +207,47 @@ describe('cookie.serialize(name, value, options)', function () {
...
 
     it('should throw for invalid value', function () {
-      assert.throws(cookie.serialize.bind(cookie, 'foo', 'bar', { path: '/\n' }),
-        /option path is invalid/)
-    })
-  })
+      const invalidPaths = [
+        '/\n',
+        '/foo\u0000',
+        '/foo bar',
```
blakeembrey, line 235: This wasn't invalid, should it be? The spec says `<any CHAR except CTLs or ";">` so I believe space is fine?

test/serialize.js
```diff
@@ -133,14 +207,47 @@ describe('cookie.serialize(name, value, options)', function () {
...
+        '/path/with\rnewline',
+        '/path\\with\\backslash',
```
blakeembrey, line 237: Same with a backslash, also fine according to `<any CHAR except CTLs or ";">`.

test/serialize.js
hdtmccallie: You are correct, good catch.

test/serialize.js
hdtmccallie: Copy that. I had it pinned with 0.12 for the install and tests but was running 8 for debug so missed this.

test/serialize.js
hdtmccallie: You are correct. Space should be acceptable.

test/serialize.js
hdtmccallie: Correct as well.

PR #175, opened 2024-10-04
Use modern JS features, ship TS defs

blakeembrey:
Refactored to modernize JS features used (`const`, `let`, etc) and bump the min node.js in prep for v1. Using TypeScript to ship native `.d.ts` files with the package moving forward. Removes some level of input validation relying on correct types (tiny perf win, less code checks).


src/index.ts
```diff
@@ -0,0 +1,370 @@
...
+ * Parse a cookie header.
+ *
+ * Parse the given cookie header string into an object
+ * The object has the various cookies as keys(names) => values
+ */
+export function parse(
+  str: string,
+  options?: ParseOptions,
+): Record<string, string> {
+  const obj: Record<string, string> = {};
```
gurgunday, line 89: Great PR!

I just have one feedback

Since this object is a map/record and can have any key, it would be beneficial and good practice to make its prototype `null`

Either by doing `const obj = Object.create(null)` or by using the more performant [NullObject method](https://github.com/delvedor/find-my-way/pull/333) (used in the Fastify router):

```js
const NullObject = function () {}
NullObject.prototype = Object.create(null)

const obj = new NullObject();
```

blakeembrey: Awesome, thanks! 

I had tried `Object.create(null)` but found the perf impact was -10% or so, but `NullObject` is comparable or better than `{}`.

pi0: Just a suggestion: I have benchmarked [this method](https://github.com/unjs/rou3/blob/main/src/_utils.ts) which had no perf penalty. (IIFE is to make sure esbuild also tree-shakes it)

src/index.ts
```diff
@@ -0,0 +1,386 @@
...
+ * CHAR              = %x01-7F
+ *                     ; defined in RFC 5234 appendix B.1
+ */
+const pathValueRegExp = /^[\u0020-\u003A\u003D-\u007E]*$/;
+
+const __toString = Object.prototype.toString;
+const __hasOwnProperty = Object.prototype.hasOwnProperty;
+
+const NullObject: any = function () {};
+NullObject.prototype = Object.create(null);
```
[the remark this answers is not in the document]
blakeembrey: Fixed in the latest commit. I'm unsure how much it helps overall since it's used in 50% of the exposed API, am I correct in understanding it's mostly for bundle size?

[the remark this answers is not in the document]
blakeembrey: The package is currently still CommonJS, so that might also waste these bytes as it won't tree shake?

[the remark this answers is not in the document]
pi0: makes sense at least until being in CommonJS

package-lock.json
gurgunday, line 1: Is there a specific reason to add the lock file?

blakeembrey: Not particularly, I usually work with it checked it so the package versions are consistent and I can use `npm ci`. I'm not strongly opinionated either way, I suppose it's possible a dependency could change their behavior and break CI for a PR 🤷 

Would you prefer to keep it omitted?

gurgunday: I don't have a strong opinion for every package but we don't use them at Fastify -- they mostly don't achieve much for consumable packages

Feel free to do as you prefer, I was just curious to see if it was intentional, sometimes people mistakenly commit it and don't realize

PR #180, opened 2024-10-07
Remove quote parsing from package

blakeembrey:
Another RFC (cc @gurgunday). Removes a single check that wouldn't be relevant if this package set the cookie to begin with. Clients do not surreptitiously add quotes to cookies (as far as I know?), it's just a cookie value is valid to be surrounded by quotes.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -124,14 +124,6 @@ export function parse(
       let valStartIdx = startIndex(str, eqIdx + 1, endIdx);
       let valEndIdx = endIndex(str, endIdx, valStartIdx);
 
-      if (
-        str.charCodeAt(valStartIdx) === 0x22 /* " */ &&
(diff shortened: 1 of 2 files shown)


gurgunday (commented):
lgtm

blakeembrey:
@gurgunday I think if we're expecting other applications (re https://github.com/jshttp/cookie/pull/179#discussion_r1791379018) it might be an issue merging this. Since another application might be using quotes. I'm still of the opinion that it's fine to shift this responsibility into `decode` though, since adding the quotes would be a job for `encode`.

gurgunday:
Yeah, just checked the RFC again, nothing about this quote removal behavior

So I agree then that this should be a decoder responsibility, and since the default encoder doesn't do quotes here, I don't see any reason to expect or even try to unquote values

PR #190, opened 2024-10-10
Update imports example in README.md

isnifer:
Since signature of exports has changed in 1.0.0 it requires to change `import` statement in modern code bases


j0k3r:
Maybe you should rebase against the master to fix the build.

But this should definitely got merged and maybe the 1.0.0 release note should mention that change as well.

Issue #65, opened 2017-04-16
maxAge is wrong


patrikx3:
Hi! Just a small issue, I think maxAge is in seconds (RFC), but jshttp/cookie uses milliseconds.
As well in NPM is shows seconds, when in the cookie underlying implementation uses milliseconds.
https://tools.ietf.org/html/rfc7234#section-5.2.2.8 - seconds
https://www.npmjs.com/package/cookie - shows seconds

But when I uses it i have to use milliseconds.
( I divided by 1000 and it was wrong, so I tried with milliseconds and it works now. :) )

dougwilson:
Yes, our API takes milliseconds, to make it easy to use with other JavaScript APIs taking milliseconds. What is written to the HTTP request is seconds, as the RFC requires, so there is no issue here.

patrikx3:
Cool, just trying to help you so the docs in NPM and README will show milliseconds instead of seconds... It is not clear right now. But thanks.

patrikx3:
ciao

dougwilson:
Oh, sorry, I didn't realize you were actually describing a doc issue, because it sounded like you were talking about the RFC. I see in https://github.com/jshttp/cookie/blob/master/README.md#maxage it is wrong, yes. Please feel free to submit a PR :)

patrikx3:
what do you mean? just fork , fix it and send you a pull??? that's what you mean?

PR #201, opened 2024-10-22
Use correct units for maxAge option

erikmusd:
See @dougwilson comment: https://github.com/jshttp/cookie/issues/65#issuecomment-294358204

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ The default function is the global `encodeURIComponent`.
 
 ##### maxAge
 
-Specifies the `number` (in seconds) to be the value for the [`Max-Age` `Set-Cookie` attribute](https://tools.ietf.org/html/rfc6265#section-5.2.2).
+Specifies the `number` (in milliseconds) to be the value for the [`Max-Age` `Set-Cookie` attribute](https://tools.ietf.org/html/rfc6265#section-5.2.2).
 
 The [cookie storage model specification](https://tools.ietf.org/html/rfc6265#section-5.3) states that if both `expires` and
 `maxAge` are set, then `maxAge` takes precedence, but it is possible not all clients by obey this,


blakeembrey:
No, the package is definitely accepting seconds and not milliseconds. It serializes the integer directly as max age. I believe there was a mistake in that comment and it was corrected later.

erikmusd:
@blakeembrey I'm using `cookie-parser` package and it [delegates the cookie options](https://github.com/expressjs/cookie-parser?tab=readme-ov-file#cookieparsersecret-options) to this package. When I used seconds it doesn't work as intended, I have to use milliseconds. So after seeing some closed issues I thought the problem was here. So maybe cookie-parser is the problem then.

blakeembrey:
There’s nothing I see in cookie parser about writing cookies, it’s probably a different package you’re using?

erikmusd:
@blakeembrey it uses `cookie `as a dependency. And accepts `cookie `options i.e. maxAge, etc.

blakeembrey:
For parsing, but not serializing. I just read over the code. You must be using something else to serialize cookies.

blakeembrey:
Maybe you're using this? https://github.com/expressjs/session?tab=readme-ov-file#cookiemaxage (it mentions using milliseconds, so that package is probably altering the input to seconds before serializing here).

erikmusd:
I'm using:

image

but idk, maybe there is something else modifying it somewhere.

blakeembrey:
That’s correct, req.cookies is the parsed cookies, it’s not for serializing them though. You have no need for maxAge when parsing.

erikmusd:
@blakeembrey my bad, I'm using the` res.cookies` (https://expressjs.com/en/api.html#res.cookie), not` req.cookies`. That's where I'm using milliseconds instead of seconds when creating a cookie. Thanks for helping clarify my misunderstanding. I missed the detail. Thanks for taking your time to check this out! Good day!

PR #207, opened 2024-11-15
fix: options.priority ternary in serialize validation

### LINKED ISSUE #206: What's the purpose of the `options.sameSite` branch of ternary in serialize?
What's the purpose of the `options.sameSite` branch of ternary in serialize?

When `options.priority` is not a string, it will use `options.sameSite`. There's no overlap in the types between the two, so we'd hit the default and throw. Was there meant to be overlap or is this a typo?
https://github.com/jshttp/cookie/blob/408c2b40c6c5783fcfae619d30e8427e9c96a2db/src/index.ts#L314
```ts
  if (options.priority) {
    const priority =
      typeof options.priority === "string"
        ? options.priority.toLowerCase()
        : options.sameSite; // typo?
    switch (priority) {
      case "low":
        str += "; Priority=Low";
        break;
      case "medium":
        str += "; Priority=Medium";
        break;
      case "high":
        str += "; Priority=High";
        break;
      default:
        throw new TypeError(`option priority is invalid: ${options.priority}`);
    }
  }
```

jonchurch:
closes #206 

Assuming here this was not intentional

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -311,7 +311,7 @@ export function serialize(
     const priority =
       typeof options.priority === "string"
         ? options.priority.toLowerCase()
-        : options.sameSite;
+        : undefined;
     switch (priority) {
       case "low":
         str += "; Priority=Low";


jonchurch:
Codecov is failing bc I opened from a fork Im guessing, so no access to the secret

blakeembrey:
Good catch, do you want to make both undefined in the ternaries? I don’t know why I didn’t do that the first time.

jonchurch:
By both you mean the next block w/

```ts
  if (options.sameSite) {
    const sameSite =
      typeof options.sameSite === "string"
        ? options.sameSite.toLowerCase()
        : options.sameSite;
```

this needs to stay, at that part of the code sameSite can be `boolean | <the valid strings>`

PR #208, opened 2024-11-15
add specific error messages to serialzie validation

jonchurch:
draft take on #163 using a custom error class, and adding in the valid type of arg when its easy to do so.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -84,6 +84,31 @@ export interface ParseOptions {
   decode?: (str: string) => string | undefined;
 }
 
+export const CookieErrorCode = {
+  INVALID_NAME: "INVALID_NAME",
+  INVALID_VALUE: "INVALID_VALUE",
+  INVALID_MAXAGE: "INVALID_MAXAGE",
+  INVALID_DOMAIN: "INVALID_DOMAIN",
+  INVALID_PATH: "INVALID_PATH",
+  INVALID_EXPIRES: "INVALID_EXPIRES",
+  INVALID_PRIORITY: "INVALID_PRIORITY",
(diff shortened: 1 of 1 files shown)


jonchurch:
Idk that the benchmark takes error creation into account, so didn't post it

jonchurch:
@MaoShizhong don't mean to step on your contribution here, started writing this before I saw your PR.

@wesleytodd I can see how error messages could be shoehorned into being considered breaking changes for a package as used as cookie, although I don't like it. The original change I was doing was just adding in the enriched error messages.

It does switch from TypeError to custom error class, which is breaking as `CookieError !== TypeError`.

The breaking parts can be backed out, and @MaoShizhong could land the error codes independently in a minor.

wesleytodd:
> I can see how error messages could be considered breaking changes for a package as widely used as cookie, aIthough I don't like it or what people would be depending on. The original change I was doing was just adding in the enriched error messages.

I would not consider this a breaking change **IF** we had provided `.code` previously. If there is a better documented way to type check an error then I am good (even if that is the less ideal `typeof` check which couples to implementation details). I think the issue here is that both are likely breaking in this PR.

Maybe you could break them apart and do a `.code` addition first?

mao-sz:
> I would not consider this a breaking change IF we had provided .code previously.

Given the above discussion, this perhaps may be a naive solution and not something I like particularly much, but would it be viable to add something like
```ts
declare global {
  interface TypeError {
    code: string;
  }
}
```
to `src/index.ts` somewhere?

Feels ugly to me but shouldn't be breaking as it'll retain the `TypeError` constructor but allow us to add `.code`. Then something like a custom class could come in a future major.

Issue #191, opened 2024-10-10
Option to bypass validation during cookie serialization


rosssteytler:
The recent updates to introduce stricter validation on cookie names and values has broken some workflows in proxy scenarios where I'm parsing, modifying and then reserializing cookies.

In this case I am not the origin of the cookies and stricter validation is not required. It would be very helpful if there was a new option to bypass the validation on name and value during serialization.

Is this something you would consider adding to the library?

blakeembrey:
I think it's possible, but it does re-introduce the vulnerability from 0.6 and below. Do you have an example of something that's failing so I can get a sense of what needs to skip validation? I'd prefer to avoid re-introducing a potential issue into the library.

rosssteytler:
In this particular case it is a cookie value that contained a comma. This was working on version 0.6 but restricted in the current version.

blakeembrey:
> In this particular case it is a cookie value that contains a comma.

Are you also trying to avoid encoding the comma? What's the decode/encode functions you use? Because this library, by default, would be escaping the comma to `%2C`.

rosssteytler:
The cookie is initially created by the origin server (not by me). In my proxy workflow, I use the cookie library to parse it, modify configurations like expires, and then reserialize it. Since I'm not the origin of the cookie, I can't alter its name or value, so I've disabled decoding and encoding. I just need the cookie to pass through my proxy as is, regardless of its validity - hence the need for an option to bypass validation in this particular scenario.

blakeembrey:
> I use the cookie library to parse it

Out of curiosity, how are you parsing it with this library? It was only designed to parse the `cookie` header so I'm guessing you had to build out some other things?

blakeembrey:
> Since I'm not the origin of the cookie, I can't alter its name or value, so I've disabled decoding and encoding.

Do you know if it's only the comma in the cookie value? I'm inclined to allow it to be a bit looser to fit what browsers support as long as it doesn't create a security issue. But bypassing everything is a little iffier due to the ability to escape field boundaries.

rosssteytler:
> Out of curiosity, how are you parsing it with this library? It was only designed to parse the `cookie` header so I'm guessing you had to build out some other things?

My proxy service receives a request intended for a target server. The proxy service then makes a fetch request to the target server and parses the raw set-cookie response headers from the target server using this library. Once I've modified certain options such as `expires` I reserialize the cookie using this library and return it in the response to the original request.

[...]

PR #210, opened 2024-11-15
Loosen cookie name/value validation

blakeembrey:
Closes https://github.com/jshttp/cookie/issues/191. All keys and values are supported in Safari. Although older or other browsers may not support some of these characters, I don't think we should be restricting it unless it's a security issue.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,8 +8,11 @@
  * tchar             = "!" / "#" / "$" / "%" / "&" / "'" /
  *                     "*" / "+" / "-" / "." / "^" / "_" /
  *                     "`" / "|" / "~" / DIGIT / ALPHA
+ *
+ * Note: Allowing more characters - https://github.com/jshttp/cookie/issues/191
+ * Allow same range as cookie value, except `=`, which delimits end of name.
  */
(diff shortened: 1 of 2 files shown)


ctcpip (changes requested):
you've actually eliminated some of what you previously allowed...

ctcpip (approved):
ignore previous comment, I was squinting wrongly at the regex

UlisesGascon (approved):

benmccann:
A couple of questions about this:
- Is 1.0.2 fully backward compatible with the pre-1.0 releases?
- Is there any need to avoid cookies with special characters such as `:` still?

blakeembrey:
> Is 1.0.2 fully backward compatible with the pre-1.0 releases?

No, the breaking changes can be viewed here: https://github.com/jshttp/cookie/releases/tag/v1.0.0

Assuming you are only referring to this PR and pre-0.7 releases, still no. The regex is stricter than what was previously causing a security issue.

> Is there any need to avoid cookies with special characters such as : still?

Cookies could contain `:`, but names were stricter following the RFC and didn't allow `:`. They now allow `:` again with this change.

PR #213, opened 2024-11-19
fix(docs): add disclaimer about parsing set-cookie

wesleytodd:
This is a common ask. Maybe this type of disclaimer might help?

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,8 @@ Parse a HTTP `Cookie` header string and returning an object of all cookie name-v
 The `str` argument is the string representing a `Cookie` header value and `options` is an
 optional object containing additional parsing options.
 
+**NOTE:** This method does not parse the `set-cookie` header. It parses the `cookie` header. This means you cannot do `cookie.parse(cookie.serialize('name', 'value'))`.
+
 ```js
 const cookies = cookie.parse("foo=bar; equation=E%3Dmc%5E2");
 // { foo: 'bar', equation: 'E=mc^2' }


bjohansebas (approved):

README.md
```diff
@@ -26,6 +26,8 @@ Parse a HTTP `Cookie` header string and returning an object of all cookie name-v
 The `str` argument is the string representing a `Cookie` header value and `options` is an
 optional object containing additional parsing options.
 
+**NOTE:** This method does not parse the `set-cookie` header. It parses the `cookie` header. This means you cannot do `cookie.parse(cookie.serialize('name', 'value'))`.
```
blakeembrey, line 29:
```suggestion
**Note:** This method does not parse the `Set-Cookie` header.
```

It already says it parses the `Cookie` header twice in the above paragraph, would prefer to keep the note shorter so someone skimming actually reads it.

blakeembrey (changes requested):

README.md
jonchurch, line 29: The example can be improved, as that code works fine.

serialize w/ no options essentially makes a valid `Cookie` string as it just sets the cookie name and value.
The real trouble (data corruption) is when folks use the options. Specifically the flag options like `HttpOnly, Secure, Partitioned` which aren't key value pairs like a cookie header.

I think we should give folks a little more info so they understand the failure state and risk.

jonchurch:
```suggestion
> ⚠️ **NOTE:** This method does not parse the `set-cookie` header. It parses the `cookie` header. This means you cannot safely parse a `set-cookie` header, or the output of `serialize`. 
>
> E.g. `cookie.parse(cookie.serialize('name', 'value', { httpOnly: true, secure: true, partitioned: true }))` will drop the `httpOnly, secure, partitioned` values entirely. Parsing `set-cookie` header values will result in unexpected results!
```

jonchurch (changes requested):

README.md
jonchurch: lmao I went in the opposite direction with my suggestion.

jonchurch: I'd split the difference

`> ⚠️ **Note:** This method does not parse the `Set-Cookie` header.`

wesleytodd: I am personally in favor of a more detailed description. Honestly I would prefer to talk about the difference and even show a detailed example (on top of what your suggestion is below). The larger we make the block warning for this the more likely someone is to see it IMO. Replying here since this is the one with the discussion, but I lean toward @jonchurch's direction below.

blakeembrey:
Closing as I don't think this is necessary with the renamed methods and updated README.

Issue #200, opened 2024-10-22
cookie.parse ignores HttpOnly; Secure;


gerhardcit:
It seems that the parses wants key=value pairs and ignores the HttpOnly; Secure; settings?

```typescript
import cookie from 'cookie';
const cookieStr = "MY_COOKIE=somevalue; Path=/; Expires=Tue, 29 Oct 2024 06:06:46 GMT; HttpOnly; Secure; SameSite=None";
const parsedCookie = cookie.parse(cookieStr, { decode: String });
console.log("parsedCookie", parsedCookie);
```

Result: (missing the HttpOnly and Secure settings?
```javascript
parsedCookie C <[Object: null prototype] {}> {
  MY_COOKIE: 'somevalue',
  Path: '/',
  Expires: 'Tue, 29 Oct 2024 06:06:46 GMT',
  SameSite: 'None'
}
```

Expected result:
```javascript
parsedCookie C <[Object: null prototype] {}> {
  MY_COOKIE: 'somevalue',
  Path: '/',
  Expires: 'Tue, 29 Oct 2024 06:06:46 GMT',
  HttpOnly: 'true',
  Secure: 'true',
  SameSite: 'None'
}
```

if the cookie value is explicit: `HttpOnly=true; Secure=true`

`const cookieStr = "MY_COOKIE=somevalue; Path=/; Expires=Tue, 29 Oct 2024 06:06:46 GMT; HttpOnly=true; Secure=true; `

then is parses correctly.

I tried 0.7.2 and 1.0.1.  Am I missing something?

gerhardcit:
#164 seems to discuss this. Does not seem that there is a consensus about what to expect?

blakeembrey:
There is a consensus, this package doesn’t parse the `Set-Cookie` header, which is what you’re trying to do. It says this in the README too, parses a `Cookie` header.

This could be added as a feature, but it’s unsupported today.

gerhardcit:
@blakeembrey , thx.. duh!, you are very correct.
my source was this:
 ```
const setCookieHeaders = res.headers.getSetCookie();
for (const setCookieHeader of setCookieHeaders) {
....
```

We use it often in workers and sveltekit projects to pass a login method to my own upstream IDP and need to pass cookies (full or partly) on to the consumer, which has another hostname.

blakeembrey:
@gerhardcit Would it be helpful if this library did support parsing the `set-cookie` header? I'm happy to consider it as a new feature in the near future.

[...]

PR #214, opened 2024-11-20
Add stringify method to mirror parse

blakeembrey:
Part of the work from https://github.com/jshttp/cookie/issues/200, adding methods that mirror the existing methods so someone can easily understand `parse`/`stringify` and (in another PR) `serialize`/`deserialize`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -90,16 +90,18 @@ export interface ParseOptions {
 }
 
 /**
+ * Cookies object.
+ */
+export type Cookies = Record<string, string | undefined>;
+
+/**
  * Parse a cookie header.
  *
  * Parse the given cookie header string into an object
  * The object has the various cookies as keys(names) => values
  */
-export function parse(
-  str: string,
-  options?: ParseOptions,
-): Record<string, string | undefined> {
-  const obj: Record<string, string | undefined> = new NullObject();
+export function parse(str: string, options?: ParseOptions): Cookies {
+  const obj: Cookies = new NullObject();
   const len = str.length;
   // RFC 6265 sec 4.1.1, RFC 2616 2.2 defines a cookie name consists of one char minimum, plus '='.
   if (len < 2) return obj;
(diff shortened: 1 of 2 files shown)


wesleytodd:
I think it would help if we had more documentation on this, maybe a description for the PR to clarify? The code looks like the goal is to take the output from `cookie.parse` and turn it back into a cookie header string? And I assume by the comment about "and in another pr seralize/desearlize" that there will be a follow up PR to add `deseralize` to parse `set-cookie` headers?

wesleytodd:
Sorry, to which question?

src/index.ts
```diff
@@ -169,7 +169,7 @@ export interface StringifyOptions {
 }
 
 /**
- * Stringify a set of cookies into a `Cookie` header string.
+ * Stringifies an object into a HTTP `Cookie` header.
```
wesleytodd, line 172:
```suggestion
 * Stringifies an object into an HTTP `Cookie` header.
```

wesleytodd (approved):

src/index.ts
```diff
@@ -155,6 +157,46 @@ function endIndex(str: string, index: number, min: number) {
...
+
+    if (!cookieNameRegExp.test(name)) {
+      throw new TypeError(`cookie name is invalid: ${name}`);
+    }
+
+    const value = enc(val);
+
+    if (!cookieValueRegExp.test(value)) {
+      throw new TypeError(`cookie val is invalid: ${val}`);
+    }
```
jonchurch, line 192: Maybe this is more about serialize, but we end up dropping something with say `{foo: undefined}` here silently.

Whereas serialize will give you `foo=undefined`

is the discrepancy intended? v0 line also serializes undefined that same way

blakeembrey: > is the discrepancy intended?

Yes, it's intended. The design of `parse` allows for `undefined` to be returned and skip that entity when parsing, so this is mirroring support that (where an object can have `undefined` as the value).

As for the serialize API, it's documented to support `string` on `serialize` and that's the defined support, anything else is out of scope for that API. We shouldn't be trying to support non-strings even if they happen to work today accidentally.

jonchurch (approved):

src/index.ts
```diff
@@ -155,6 +157,46 @@ function endIndex(str: string, index: number, min: number) {
...
+ * Stringifies an object into a HTTP `Cookie` header.
+ */
+export function stringify(
+  cookies: Cookies,
+  options?: StringifyOptions,
+): string {
+  const enc = options?.encode || encodeURIComponent;
+  const cookieStrings: string[] = [];
+
+  for (const [name, val] of Object.entries(cookies)) {
```
gurgunday, line 181: This might be slower than `Object.keys` and accessing the object as it creates many arrays

PR #215, opened 2024-12-06
Fix TS inconsistency (parse method never returns undefined).

davidesigner:
In this PR: #179, `undefined` was added to the type definition and code, then the code was refactored but the type definition was not updated. See the review here: https://github.com/jshttp/cookie/pull/179#discussion_r1791199926 to understand the history.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -86,7 +86,7 @@ export interface ParseOptions {
    *
    * @default decode
    */
-  decode?: (str: string) => string | undefined;
+  decode?: (str: string) => string;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


blakeembrey:
Thanks for the PR, but it's correct that `decode` can return `undefined` and trying to remove it now would be a breaking change. What you are referring to is the default behavior, which I left designed to mimic 0.x, but the type definition does allow an application to specifically set the decode behavior I wrote in the PR initially in case they wanted to discard values they don't understand.

This is much more relevant when people might want to replace decoding with e.g. `base64url`.

PR #223, opened 2025-03-13
refactor: optimize serialization function with modern JavaScript features

Ayoub-Mabrouk:
refactor: optimize serialization function with modern JavaScript features

- Replaced string concatenation with template literals for better readability.
- Simplified option validation using object maps for priority and sameSite.
- Improved type safety with `instanceof Date` check for `expires` and added const assertions.
- Enhanced sameSite handling for boolean values.
- Removed unused isDate helper.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -250,107 +250,80 @@ export function serialize(
   val: string,
   options?: SerializeOptions,
 ): string {
-  const enc = options?.encode || encodeURIComponent;
+  const enc = options?.encode ?? encodeURIComponent;
 
   if (!cookieNameRegExp.test(name)) {
     throw new TypeError(`argument name is invalid: ${name}`);
   }
 
   const value = enc(val);
-
   if (!cookieValueRegExp.test(value)) {
     throw new TypeError(`argument val is invalid: ${val}`);
   }
 
-  let str = name + "=" + value;
+  let str = `${name}=${value}`;
   if (!options) return str;
 
   if (options.maxAge !== undefined) {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -250,107 +250,80 @@ export function serialize(
...
     }
-
-    str += "; Path=" + options.path;
+    str += `; Path=${options.path}`;
   }
 
   if (options.expires) {
     if (
-      !isDate(options.expires) ||
+      !(options.expires instanceof Date) ||
```
blakeembrey, line 290: This would be a breaking change since a `Date` could currently be passed cross frame.

blakeembrey (changes requested):
I don't think this change is currently needed. I did some perf testing earlier and for whatever reason changing these to string interpolation was negative. It really shouldn't be but that's what happened.

src/index.ts
```diff
@@ -250,107 +250,80 @@ export function serialize(
...
+    str += `; Priority=${priority}`;
   }
 
   if (options.sameSite) {
+    const sameSiteMap = {
+      strict: "Strict",
+      lax: "Lax",
+      none: "None",
+      true: "Strict",
+    } as const;
```
blakeembrey, line 319: I don't think replacing the current code is necessary, this could just as easily be changed instead to use the string or lower cased string directly. We don't specifically need it to be capitalized first.

That said though, if you were to make a change like this the object shouldn't be generated inline every time and you'd want to set it once as a global.

PR #253, opened 2025-11-26
Overwrite value in passed in options

blakeembrey:
From X, a user reported a regression and it looks like an unfortunate conflict in property names due to `options` backward-compatibility. 

Issue in code: https://github.com/elysiajs/elysia/blob/5eec011e53784445dc9d63635aeccb4f4d5d5c3e/src/cookies.ts#L445-L453

Caused by `value` being passed as an argument and also in `options`, and `options` took priority instead of the expected `value`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -293,7 +293,7 @@ export function stringifySetCookie(
   const cookie =
     typeof _name === "object"
       ? _name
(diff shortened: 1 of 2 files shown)


brandonmcconnell (commented):
Thanks for the quick fix here, @blakeembrey!

PR #255, opened 2025-12-28
README: fix docs around parse() / stringify() function names

getify:
One fix here is an inconsistency where `stringify()` in code example hadn't been updated to the new `stringifyCookie()` name of the section the example was in.

The other fix is adding a note to each section (`parseCookie()` and `stringifyCookie()`) to indicate their now-deprecated alias names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ const cookie = require("cookie");
 
 ### cookie.parseCookie(str, options)
 
+Alias: `parse()` (deprecated)
+
 Parse an HTTP `Cookie` header string and return an [object](#cookie-object) of all cookie name-value pairs.
 The `str` argument is the string representing a `Cookie` header value and `options` is an
 optional object containing additional parsing options.
@@ -36,6 +38,8 @@ const cookieObject = cookie.parseCookie("foo=bar; equation=E%3Dmc%5E2");
 - `decode` Specifies the function to decode a [cookie-value](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1). Defaults to [`decodeURIComponent`](#encode-and-decode).
 
 ### cookie.stringifyCookie(cookieObj, options)
+
+Alias: `stringify()` (deprecated)
 
 Stringifies a [cookie object](#cookie-object) into an HTTP `Cookie` header.
 
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ const cookie = require("cookie");
 
 ### cookie.parseCookie(str, options)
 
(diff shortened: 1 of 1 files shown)


bjohansebas (approved):

README.md
```diff
@@ -22,6 +22,8 @@ const cookie = require("cookie");
 
 ### cookie.parseCookie(str, options)
 
+Alias: `parse()` (deprecated)
```
blakeembrey: Can you share more context around adding the deprecated names? I can guess it’s for quick lookup from a previous version, but would like to understand more. 

WDYT about making a “deprecated” section at the bottom and using that instead to point to the new methods? It could contain more information about the old function format too (three args vs one for stringify).

getify: I upgraded from version 1.0.2 to 1.1.1, and as I was checking the docs, was confused by the new names... it felt at first like it had been a breaking change, even though it was only a minor version update.

I didn't see that the old names were still there for several minutes (where I was searching through the commit history to find out when that "breaking change" had happened), and then found the aliases were included down at the bottom of the file.

So my goal in adding them to the documentation is to prevent others' confusion. I made the "assumption" of deprecation given the name change but if there's no intent to actually deprecate, that part could be taken out from the names.

If you're intending to deprecate (and eventually break, on a semver major) the arguments, I agree those changes should also be noted in the docs, and indeed maybe a "Deprecation" section is appropriate.

But even still, I think to prevent confusion for others, like I had, I think having some note like my "Alias: .." in each section is helpful (at least for the rest of the 1.x branch) so people upgrading can see that they used to call `parse()` and can still do so, but *should* update to the `parseCookie()` name, or whatever).

If you have other thoughts on how best to document, I'm happy to update the PR... or you can do so. Either way, was just trying to help others not hit my confusion. :)

README.md
```diff
@@ -39,8 +39,6 @@ const cookieObject = cookie.parseCookie("foo=bar; equation=E%3Dmc%5E2");
 
 ### cookie.stringifyCookie(cookieObj, options)
 
-Alias: `stringify()` (deprecated)
```
blakeembrey, line 42: Needed to move this, it was `serialize` and the new method is `stringifySetCookie` (common mistake and why it was renamed when there's now two different stringify methods).

README.md
```diff
@@ -242,6 +242,14 @@ function onRequest(req, res) {
 http.createServer(onRequest).listen(3000);
 ```
 
+## Deprecated
```
blakeembrey, line 245: I added a lightweight deprecated section and linked it, there's not a lot to include but it can be expanded in the future.

PR #260, opened 2026-01-25
refactor: use includes() instead of indexOf() for boolean checks

Ayoub-Mabrouk:
Replace str.indexOf("%") === -1 with !str.includes("%") in decode function. This improves readability and may have slight performance benefits as includes() is optimized for boolean checks in modern JavaScript engines.


blakeembrey:
While this seems true of V8, it looks like Safari results in a perf hit using includes. I'm not sure why, but it's 75% slower than `indexOf`. Since this library can be used isomorphically I don't think I'll be merging this right now.

Issue #250, opened 2025-10-05
Performance optimization opportunities for cookie handling


jdmiranda:
## Summary

The `cookie` package is a critical dependency in the Node.js ecosystem, used on **every HTTP request** that involves cookies. It serves as the foundation for Express session management, authentication, and other cookie-based functionality. Given its position as a hot path in request processing, even small optimizations can have significant impact across the ecosystem.

I've identified several performance optimization opportunities that could provide meaningful gains, especially for high-traffic applications.

## Performance Context

- **Call frequency**: Executed on every request with cookies (sessions, auth, tracking, etc.)
- **Critical path**: Part of request parsing before application logic runs
- **Ecosystem impact**: Used by Express, Koa, and countless other frameworks
- **Current dependency count**: ~4,794 packages depend on this library

### 1. RegExp Validation Caching

**Current behavior**: Every call to `stringifySetCookie` re-runs RegExp tests on the same cookie names and attribute values, even when they're identical across requests.

**Opportunity**: Cache validation results for commonly-used cookie names and values.

```typescript
// Add a validation cache
const VALIDATION_CACHE = {
  names: new Map<string, boolean>(),
  values: new Map<string, boolean>(),
  domains: new Map<string, boolean>(),
  paths: new Map<string, boolean>(),
};

// Cache size limits to prevent memory issues
const MAX_CACHE_SIZE = 100;

function isValidCookieName(name: string): boolean {
  let valid = VALIDATION_CACHE.names.get(name);
  if (valid !== undefined) return valid;
  
  valid = cookieNameRegExp.test(name);
  if (VALIDATION_CACHE.names.size < MAX_CACHE_SIZE) {
    VALIDATION_CACHE.names.set(name, valid);
  }
  return valid;
}
```

**Impact**: 15-25% improvement for serialization operations with common cookie names (session_id, auth_token, csrf_token, etc.)

### 2. Date.toUTCString() Caching for Common Expiration Times

**Current behavior**: `cookie.expires.toUTCString()` is called repeatedly for the same expiration timestamps, particularly for session cookies with standard durations.

**Opportunity**: Cache formatted date strings for recently-used timestamps.

```typescript
// LRU-style cache for date strings
const DATE_CACHE_SIZE = 50;
const dateStringCache = new Map<number, string>();

function getUTCString(date: Date): string {
  const timestamp = date.getTime();
  
  // Check cache
  let cached = dateStringCache.get(timestamp);
  if (cached) return cached;
  
[...]

PR #262, opened 2026-02-13
Faster stringify cookie with benchmarks

blakeembrey:
I decided to try investigating if anything in https://github.com/jshttp/cookie/issues/250 holds true, and found the opposite for string concatenation. Updated to use string concatenation to improve performance.

Before: 

```
 ✓ src/stringify-cookie.bench.ts > cookie.stringifyCookie 7195ms
     name                         hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · empty             19,138,902.74  0.0000  0.0572  0.0001  0.0000  0.0001  0.0001  0.0002  ±0.11%  9569452
   · simple             5,777,350.07  0.0001  0.3515  0.0002  0.0002  0.0002  0.0003  0.0006  ±0.39%  2888676
   · undefined values   3,129,664.63  0.0002  0.3241  0.0003  0.0003  0.0004  0.0004  0.0008  ±0.46%  1564833
   · 10 cookies           765,737.48  0.0012  0.4257  0.0013  0.0013  0.0014  0.0016  0.0020  ±0.34%   382869
   · 100 cookies           78,240.71  0.0122  0.4788  0.0128  0.0128  0.0137  0.0160  0.0424  ±0.39%    39121
```

After:

```
 ✓ src/stringify-cookie.bench.ts > cookie.stringifyCookie 7958ms
     name                         hz     min     max    mean     p75     p99    p995    p999     rme   samples
   · empty             21,654,747.61  0.0000  0.1649  0.0000  0.0000  0.0001  0.0001  0.0002  ±0.14%  10827374
   · simple             7,345,985.72  0.0000  0.0540  0.0001  0.0001  0.0002  0.0002  0.0003  ±0.10%   3672994
   · undefined values   3,820,139.53  0.0002  0.0491  0.0003  0.0003  0.0003  0.0004  0.0004  ±0.11%   1910070
   · 10 cookies           854,240.43  0.0010  0.4322  0.0012  0.0012  0.0013  0.0014  0.0019  ±0.44%    427121
   · 100 cookies           85,417.71  0.0110  0.4168  0.0117  0.0117  0.0126  0.0160  0.0470  ±0.44%     42709
```

### DIFF
--- a/src/stringify-cookie.bench.ts
+++ b/src/stringify-cookie.bench.ts
@@ -1,0 +1,41 @@
+import { describe, bench } from "vitest";
+import * as cookie from "./index.js";
+
+describe("cookie.stringifyCookie", () => {
+  bench("empty", () => {
+    cookie.stringifyCookie({});
+  });
+
+  bench("simple", () => {
+    cookie.stringifyCookie({ foo: "bar" });
+  });
+
+  bench("undefined values", () => {
+    cookie.stringifyCookie({
+      foo: "bar",
+      baz: undefined,
+      qux: "quux",
+      zap: undefined,
+    });
+  });
+
+  const cookies10 = genCookies(10);
+  bench("10 cookies", () => {
+    cookie.stringifyCookie(cookies10);
+  });
+
+  const cookies100 = genCookies(100);
+  bench("100 cookies", () => {
+    cookie.stringifyCookie(cookies100);
+  });
+});
+
+function genCookies(num: number) {
+  const cookies: Record<string, string | undefined> = {};
+
+  for (let i = 0; i < num; i++) {
+    cookies["foo" + i] = "bar" + i;
+  }
+
+  return cookies;
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -158,9 +158,11 @@ export function stringifyCookie(
   options?: StringifyOptions,
 ): string {
   const enc = options?.encode || encodeURIComponent;
-  const cookieStrings: string[] = [];
+  const keys = Object.keys(cookie);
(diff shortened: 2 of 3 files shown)


bjohansebas (approved):

src/stringify-cookie.bench.ts
```diff
@@ -0,0 +1,41 @@
+import { describe, bench } from "vitest";
```
sheplu, line 1: would that make more sense to move the two new bench.ts file in the benchmark folder already existing?

sheplu (changes requested):
Just a change for the folder that should host the two new bench files 
Other than than LGTM!

src/stringify-cookie.bench.ts
blakeembrey: No, I prefer to collocate relevant tests and benchmarks but these new benchmarks don’t make sense there since that folders more for general scripts. These are specifically benchmarking the functions next to them, while the benchmarks directory does not import anything from the source files.

blakeembrey: Can you unblock the PR? Another way to think about the structure is:

1. `benchmarks` are for independent code, they can be run independently and are more saved curiosities (closer to ADRs than tests, they show why certain approaches were selected)
2. Co-located benchmarks are for the current code, they cannot be run independently and are only relevant when comparing to a previous code state to know changes over time (more similar to the tests)

sheplu: all good thanks for the structure!

sheplu (approved):
LGTM

PR #269, opened 2026-04-28
Optimize stringifyCookie by 2x for cookie-octet values

saripovdenis:
# `stringifyCookie` Optimization

Optimizes default cookie encoding by skipping `encodeURIComponent` when values are already roundtrip-safe cookie-octets.

`%` is still encoded to preserve roundtrip behavior:

```ts
stringifyCookie({ foo: "%20" }) // foo=%2520
parseCookie("foo=%2520")        // { foo: "%20" }
```

## Benchmarks

> Higher `hz` is better. Measured on Node 24.15.0.

| case | before hz | after hz | speedup |
|---|---:|---:|---:|
| `empty` | 38,598,774.76 | 40,958,547.67 | 1.06x |
| `simple` | 11,020,607.14 | 16,615,825.47 | 1.51x |
| `rfc cookie-octets` | 7,463,520.33 | 15,267,614.96 | 2.05x |
| `encode` | 8,424,626.32 | 7,635,768.40 | 0.91x |
| `undefined values` | 5,783,315.75 | 8,068,459.56 | 1.40x |
| `mixed encode` | 3,977,455.71 | 4,320,201.87 | 1.09x |
| `10 cookies` | 1,297,376.60 | 1,895,381.87 | 1.46x |
| `100 cookies` | 130,578.34 | 179,100.23 | 1.37x |

## Summary

- `rfc cookie-octets`: **2.05x faster**
- `simple`: **1.51x faster**
- `10 cookies`: **1.46x faster**
- `100 cookies`: **1.37x faster**
- `encode` case: **0.91x**, slower because it checks the fast-path regexp before falling back to `encodeURIComponent`

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ const cookieHeader = cookie.stringifyCookie({ a: "foo", b: "bar" });
 
 #### Options
 
-- `encode` Specifies the function to encode a [cookie-value](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1). Defaults to preserving roundtrip-safe cookie-octet values and using [`encodeURIComponent`](#encode-and-decode) otherwise.
+- `encode` Specifies the function to encode a [cookie-value](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1). Defaults to [`encodeURIComponent`](#encode-and-decode).
 
 ### cookie.parseSetCookie(str, options)
 
@@ -88,7 +88,7 @@ const setCookieHeader = cookie.stringifySetCookie({
 
 #### Options
 
-- `encode` Specifies the function to encode a [cookie-value](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1). Defaults to preserving roundtrip-safe cookie-octet values and using [`encodeURIComponent`](#encode-and-decode) otherwise.
+- `encode` Specifies the function to encode a [cookie-value](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1). Defaults to [`encodeURIComponent`](#encode-and-decode).
 
 ## Cookie object
 
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ const cookieHeader = cookie.stringifyCookie({ a: "foo", b: "bar" });
 
 #### Options
 
-- `encode` Specifies the function to encode a [cookie-value](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1). Defaults to preserving valid cookie-octet values and using [`encodeURIComponent`](#encode-and-decode) otherwise.
+- `encode` Specifies the function to encode a [cookie-value](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1). Defaults to preserving roundtrip-safe cookie-octet values and using [`encodeURIComponent`](#encode-and-decode) otherwise.
 
 ### cookie.parseSetCookie(str, options)
 
@@ -88,7 +88,7 @@ const setCookieHeader = cookie.stringifySetCookie({
 
 #### Options
 
-- `encode` Specifies the function to encode a [cookie-value](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1). Defaults to preserving valid cookie-octet values and using [`encodeURIComponent`](#encode-and-decode) otherwise.
+- `encode` Specifies the function to encode a [cookie-value](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1). Defaults to preserving roundtrip-safe cookie-octet values and using [`encodeURIComponent`](#encode-and-decode) otherwise.
 
 ## Cookie object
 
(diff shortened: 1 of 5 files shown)


saripovdenis:
@blakeembrey what do you think about this tradeoff?

blakeembrey:
This looks pretty great! I think it's worth the trade-off.

One question, should we just allow all RFC valid values through without encoding? It may differ slightly from `encodeURIComponent` (haven't checked) but might allow more values to skip encoding altogether without any regression in behavior.

saripovdenis:
> This looks pretty great! I think it's worth the trade-off.
> 
> One question, should we just allow all RFC valid values through without encoding? It may differ slightly from `encodeURIComponent` (haven't checked) but might allow more values to skip encoding altogether without any regression in behavior.

Great idea!

Some cases would have different behaviour
```ts
stringifyCookie({ foo: "a=b" })
// before: foo=a%3Db
// after:  foo=a=b
```

But overall looks like a great win from perf side!

I pushed code accordingly

saripovdenis:
Hi @blakeembrey 

Friendly ping~
Anything I can do to move this PR forward?

blakeembrey:
There's an existing regex `cookieValueRegExp`, and the difference is `"` and `,` support. Should we use the same regex or do you think those need to be encoded? And if we think they _must_ be encoded, should we change `cookieValueRegExp` instead to be stricter? Otherwise a custom encoding can still use those values.

blakeembrey:
Also I'm planning to land this in a new major version to eliminate concerns around breaking changes.

If you want this in the current version I can cherry pick your original commit, just let me know.

src/stringify-cookie.bench.ts
```diff
@@ -10,6 +10,14 @@ describe("cookie.stringifyCookie", () => {
     cookie.stringifyCookie({ foo: "bar" });
   });
 
+  bench("rfc cookie-octets", () => {
+    cookie.stringifyCookie({ foo: "a=b+c/d?x%20" });
```
blakeembrey: Actually, we'd 100% need to encode `%` to avoid ambiguity here. So I think the fast path is `cookieValueRegexp` minus `%`?

Otherwise the round trip won't be correct.

src/index.ts
```diff
@@ -68,6 +68,11 @@ const pathValueRegExp = /^[\u0020-\u003A\u003D-\u007E]*$/;
  */
 const maxAgeRegExp = /^-?\d+$/;
 
+/**
+ * RegExp to match RFC 6265 cookie-octet values that need no URL encoding.
+ */
+const cookieOctetRegExp = /^[!#$%&'()*+\-.\/0-9:<=>?@A-Z[\]\^_`a-z{|}~]*$/;
```
blakeembrey:
```suggestion
const cookieOctetRegExp = /^[!#$&'()*+\-.\/0-9:<=>?@A-Z[\]\^_`a-z{|}~]*$/;
```

blakeembrey: Make sure to add a test that `%20` should round trip properly, e.g. `decode(encode('%20'))` should work.

saripovdenis:
> There's an existing regex `cookieValueRegExp`, and the difference is `"` and `,` support. Should we use the same regex or do you think those need to be encoded? And if we think they _must_ be encoded, should we change `cookieValueRegExp` instead to be stricter? Otherwise a custom encoding can still use those values.

In vacuum I would make `cookieValueRegExp` stricter, but I think you had a spicy issue here haha
https://github.com/jshttp/cookie/issues/191

so I would keep cookieValueRegExp the same as now to be more permissive for people who need it

src/index.ts
saripovdenis: Very good catch, done

thank you!

README.md
```diff
@@ -48,7 +48,7 @@ const cookieHeader = cookie.stringifyCookie({ a: "foo", b: "bar" });
 
 #### Options
 
-- `encode` Specifies the function to encode a [cookie-value](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1). Defaults to [`encodeURIComponent`](#encode-and-decode).
+- `encode` Specifies the function to encode a [cookie-value](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1). Defaults to preserving roundtrip-safe cookie-octet values and using [`encodeURIComponent`](#encode-and-decode) otherwise.
```
blakeembrey: I would leave these unchanged, and keep the extra detail in the "encode and decode" docs.

src/stringify-cookie.spec.ts
```diff
@@ -28,12 +28,60 @@ describe("cookie.stringifyCookie", () => {
...
+      "^_`abcdefghijklmnopqrstuvwxyz{|}~";
+
+    expect(stringifyCookie({ foo: value })).toEqual(`foo=${value}`);
+  });
+
+  it("should match default encoding for all valid BMP characters", () => {
+    const cookieOctets =
+      "!#$&'()*+-./0123456789:<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[]" +
+      "^_`abcdefghijklmnopqrstuvwxyz{|}~";
+    const mismatches: string[] = [];
```
blakeembrey: I would try to use the testing framework for this, it's harder to debug `expect(mismatches).toEqual([])` than a proper string comparison.

To test on multiple inputs, you can generate the list of things you want to test outside function and use `it.each()` to iterate over, allowing you to use `expect(str).toEqual(expected)` but making the inputs dynamic.

saripovdenis: Thanks!

README.md
saripovdenis: Thanks!

saripovdenis:
@blakeembrey cleaned up tests and readme!

saripovdenis:
Hey @blakeembrey, do you think we can merge?

saripovdenis:
ping~ @blakeembrey

PR #270, opened 2026-04-28
doc: Documents existing `parseCookie` behavior

saripovdenis:
## Summary

Documents existing `parseCookie` behavior when a custom `decode` returns `undefined`.

When `decode` returns `undefined`, `parseCookie` skips that value. If the same cookie name appears later, the later value can be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ const cookieObject = cookie.parseCookie("foo=bar; equation=E%3Dmc%5E2");
 
 #### Options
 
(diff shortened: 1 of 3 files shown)


blakeembrey:
Thanks for the PR, but this a vague undocumented feature that allows someone to skip a value in favor of a second one. It should have test coverage, would you be interested in adding that instead?

blakeembrey:
We can also document `undefined` behavior, I think that's preferable to total correctness and a perf hit in the fast path.

saripovdenis:
> We can also document `undefined` behavior, I think that's preferable to total correctness and a perf hit in the fast path.

oops, added it back and documented!

PR #271, opened 2026-04-28
Fix parse set cookie invalid attributes

saripovdenis:
## Summary

Avoid returning unusable values for malformed `Set-Cookie` attributes.

## Changes

- Ignore `Domain` when no value is provided.
- Avoid returning `maxAge: Infinity` when a digit-only `Max-Age` value overflows JavaScript's finite number range.

## Why

RFC 6265 says a `Set-Cookie` attribute without `=` has an empty attribute value:

https://datatracker.ietf.org/doc/html/rfc6265#section-5.2

For `Domain`, an empty value should be ignored entirely:

https://datatracker.ietf.org/doc/html/rfc6265#section-5.2.3

For `Max-Age`, digit-only values can still overflow in JavaScript:

```ts
Number("9".repeat(400)) === Infinity
```

This parser should not return `maxAge: Infinity`; it should treat non-finite parsed values as invalid and ignore the attribute.

## Tests

Added coverage for:

- `Max-Age` values that overflow to `Infinity`
- `Domain` without a value

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -443,7 +443,7 @@ export function parseSetCookie(str: string, options?: ParseOptions): SetCookie {
         if (val !== undefined) setCookie.domain = val;
         break;
       case "path":
-        if (val !== undefined) setCookie.path = val;
+        setCookie.path = val;
         break;
       case "max-age":
         if (val && maxAgeRegExp.test(val)) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -440,13 +440,16 @@ export function parseSetCookie(str: string, options?: ParseOptions): SetCookie {
         setCookie.partitioned = true;
         break;
       case "domain":
-        setCookie.domain = val;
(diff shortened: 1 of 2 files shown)


blakeembrey:
I'm not sure either of these changes are worth doing tbh:

1. `domain = undefined` and `domain` not being on the object is effectively the same behavior, and if anything it's useful to indicate there was an attribute unset.
2. Max age infinity is similar, but I could go for it. Perhaps updating the regex might be simpler to limit the max length?

saripovdenis:
> I'm not sure either of these changes are worth doing tbh:
> 
> 1. `domain = undefined` and `domain` not being on the object is effectively the same behavior, and if anything it's useful to indicate there was an attribute unset.
> 2. Max age infinity is similar, but I could go for it. Perhaps updating the regex might be simpler to limit the max length?

got it, then closing this MR

PR #273, opened 2026-05-07
Switch to ESM

blakeembrey:
For the next major version will go ESM only, min node 22 (currently in maintenance).

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
     "Douglas Christopher Wilson <doug@somethingdoug.com>"
   ],
   "sideEffects": false,
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "type": "module",
+  "exports": "./dist/index.js",
   "files": [
     "dist/"
   ],
@@ -42,7 +42,7 @@
     "vitest": "^3.2.4"
   },
   "engines": {
-    "node": ">=18"
+    "node": ">=22"
   },
   "size-limit": [
     {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,9 +16,8 @@ jobs:
     strategy:
       matrix:
         node-version:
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -17,8 +17,8 @@
     "Douglas Christopher Wilson <doug@somethingdoug.com>"
   ],
   "sideEffects": false,
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "type": "module",
```
blakeembrey, line 20: TypeScript automatically reads this and emits ESM (based on `module: nodenext`).

bjohansebas (approved):

package.json
```diff
@@ -42,7 +42,7 @@
     "vitest": "^3.2.4"
   },
   "engines": {
-    "node": ">=18"
+    "node": ">=22"
```
bjohansebas, line 45: Does 22.0.0 already support `require(esm)`?

blakeembrey: Yes: https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require

But switching to ESM isn't dependent on node 22, we could support a lower version since ESM has been supported longer. But if the concern is availability in Express 6, it will be usable 👍

blakeembrey: We _could_ even go for node 24 tbh, let me check with the TC. Worst case it can be bumped lower in the future, but we can't bump it higher without a major.

bjohansebas: Yeah, I mean, I assume Express will use versions that support `require(esm)`. In any case, support for these packages shouldn't be too tightly coupled to Express. If they want to support older Node.js versions, that should be fine as long as it doesn't become a blocker for Node.js going forward. For example, `cookie` doesn't really have anything that would block Node.js—it’s a lightweight package and isn't particularly dependent on Node.js itself or on a specific runtime.

PR #277, opened 2026-06-23
fix: apply the encode callback to empty cookie values

spokodev:
The documented `encode` option is meant to mirror `decode` and apply to every value, so that a custom encoder (signing, encrypting, base64, etc.) round-trips correctly with its matching `decode`.

`stringifySetCookie` instead used a truthiness test:

```ts
const value = rawValue ? enc(rawValue) : "";
```

For an empty-string value this skips the encoder entirely and emits a raw `name=`, instead of `name=<encode("")>`. That silently breaks encode/decode symmetry whenever the value happens to be empty.

The sibling `stringifyCookie` helper always calls the encoder, and so did `cookie@0.7.2`, so this is also an inconsistency between the two stringify paths.

The fix swaps the truthiness check for a null check:

```ts
const value = rawValue != null ? enc(rawValue) : "";
```

Empty strings are now routed through the encoder, while the object-form `value: undefined` still maps to an empty value, preserving existing behavior.

Added a test that fails before the change and passes after:

```ts
it("should apply the encoder to an empty value", function () {
  expect(
    cookie.stringifySetCookie("foo", "", { encode: (v) => "[" + v + "]" }),
  ).toEqual("foo=[]");
});
```

Full suite stays green (230 tests).


blakeembrey:
Since the documentation states it should be a string, I think this should always just be `enc(rawValue)`. However, I don't think it's worth changing except in a major version. I can't recall why I made it skip `encode`, it was probably a premature performance optimization.

spokodev:
Agreed that always `enc(rawValue)` is the cleaner contract. My change is narrower and non-breaking: it only routes the empty-string value through `encode` like every other value, so a custom `encode` sees consistent behavior without a major bump. If you would rather fold this into the broader `enc(rawValue)` change in the next major, I am fine closing this in favor of that. Your call on timing.

blakeembrey:
I'd just fold it into a new major, that's fine by me. No one has mentioned it as an issue and explicitly unsetting a cookie is usually done by passing an empty string as the value, so making it potentially longer isn't much of a plus 🤷

spokodev:
Fair enough - the empty value is usually an unset, so leaving it unencoded is the lighter default. I will close this in favor of folding it into the next major. Thanks for thinking it through.

blakeembrey:
Would you like to update this PR or make another? A new major will occur this or next week.

blakeembrey:
😅 It looks like I didn't realize my types do allow `undefined`, so LGTM, I merged with master and it'll be good to land for the next major.

PR #281, opened 2026-06-30
chore: do not include sourcemap in dist

florian-lefebvre:
`cookie` is [pretty big compared to alternatives](https://npmx.dev/compare?packages=cookie,cookie-es,js-cookie,es-cookie&facets=installSize,totalDependencies,downloads,lastUpdated,githubStars,engines,types,moduleFormat), and part of it is due shipping a sourcemap. This is not necessary so it removes it (size: from 60.8 kB to 34kB). This should likely be a patch release.

Made as part of the [e18e](https://e18e.dev/) intiative.

### DIFF
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -1,4 +1,8 @@
 {
   "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "sourceMap": false,
+    "inlineSources": false
+  },
   "exclude": ["src/**/*.spec.ts", "src/**/*.bench.ts"]
 }


blakeembrey:
Is there a good reason to do this? Shaving a few kb off the download size seems like a vanity metric, does it actually improve anyone's DX? Source maps could be the reason someone enjoys an improved DX when debugging an issue in development or production.

If we're purely trying to minimize download size, why not also ship minified code?

gameroman:
> Is there a good reason to do this? Shaving a few kb off the download size seems like a vanity metric, does it actually improve anyone's DX? Source maps could be the reason someone enjoys an improved DX when debugging an issue in development or production.
> 
> If we're purely trying to minimize download size, why not also ship minified code?

Sourcemaps currently take almost 50% of the package size

If the distributed code is readable, sourcemaps don't change much for debugging

And most consumers won't be debugging anyway,
and if you're you'll want to use the actual source code anyway

florian-lefebvre:
Sorry I should have made the reasoning clearer. Shipping sourcemaps is only useful if your code is not readable, which here is not the case (so it's not useful for debugging because the dist/index.js is already good).

Now on the topic of minification, consider these cases:
1. You ship readable code without sourcemaps (easy to debug, quite small)
2. You ship minified code with sourcemaps (hard to debug in node_modules, quite big. sourcemap currently is about 3x bigger than the source)

It's just better to ship code unbundled.

When it comes to size, it's just not for vanity. With 134M weekly downloads, it has big repercussions in bandwidth across the ecosystem: 8.15 TB to 4.56 TB (3.59 TB saved)

EDIT: didn't see your comment Roman!

blakeembrey:
So I just checked the packed size with and without, and it's more like 13.6 kB vs 9.5 kB:

```
npm notice package size: 13.6 kB
npm notice unpacked size: 64.1 kB
```

```
npm notice package size: 9.5 kB
npm notice unpacked size: 34.9 kB
```

So zipped is much less meaningful than you propose. We're not saving 3TB. The overhead of the source maps is less than half the uncompressed README. That's a saving of ~0.5TB, which at AWS egress pricing is < $50/month. Obviously not nothing, but I'm reticent to alter something that could help someone debug and for that person it _might_ be worth $50.

You are right that it is reasonably readable already, but that might not always be true depending on transpiler configuration. I don't have a good argument either direction tbh, but to me the 4kb saving seems negligible and I worry for an ecosystem that's trying to shave a kb here or there to be "better" than another package.

blakeembrey:
Honestly with the README size there might be savings there too, and this is then a kind of rabbit hole of optimization. E.g. if I remove just benchmarking, test, and installation instructions it's 1.6kb already. Arguably people aren't reading the README in their package so we could go even further and just make it a link to GitHub, saving a full 3kb.

I don't really want to be maintaining things where I'm trying to optimize this. I already heavily optimize my code for bundle size, which does affect real users, and may be more impactful on bandwidth than the source maps if it's distributed more than it's downloaded.

I'll think about it, but it's not an obvious win to me and opens the door to stuff I don't really want to spend my days thinking about, which is much more important to me nowadays as a maintainer than a few kb.

43081j:
> but that might not always be true depending on transpiler configuration

If you or another maintainer ever makes it not readable, I'd call that a regression - one that needs fixing.

Today, these 3 things are shipped:

- Types
- Sourcemaps
- Code

The code is just transpiled TypeScript, which means it is pretty much identical to what's in this repo minus the type annotations.

If I wanted those type annotations, that's exactly what the types are for.

That means the only person who would need these sourcemaps then must be someone who:

- debugs their `node_modules`
- wants to see the types in devtools (or similar debugger)

Given this package has ~130M downloads/week, I'm skeptical of if the majority of those users fit that profile. I feel like an incredibly tiny minority, if anyone, fit that profile and everyone else pays the price.

Side notes: the README stuff seems like a bit of a tangent, I would forget about it. Disk size and over-the-wire size matter, and removing sourcemaps will reduce both of those. Being one small part of a wider dependency tree, these effects add up and certainly do result in large savings in the long run.

blakeembrey:
> I'm skeptical of if the majority of those users fit that profile.

100% agreed, but it's a 4kb cost. And the maintenance overhead.

> If you or another maintainer ever makes it not readable, I'd call that a regression

Are you planning to add tooling for this or just hope?

I'll bring it up with the full TC and see if there's consensus but I really don't want to open this can of worms where I'm maintaining things to trim a kb of download here and there. It's a slippery slope, not just effort but enjoying actually being a maintainer, and this should have merit aside from saving a few kb otherwise what's the argument that we shouldn't be trimming the README, LICENSE, or compressing the JS further?

43081j:
> 100% agreed, but it's a 4kb cost. And the maintenance overhead.

Can you explain what the maintenance overhead is? Think I'm missing some context as it should just be a flag. 

On a side note, no not a 4kb cost. A 30KB cost. We can't forget disk cost. As mentioned above, this is two values: over the wire, and on disk. Both will reduce if we do this. 

This package is one small piece of a much larger puzzle. If 100 of those packages save 30KB of disk, that's not an insignificant amount. The same for over the wire cost. The picture is bigger than this repo. 

We should reverse this a little. Why do you _want_ the source maps? We now know removing them will be a saving for everyone, and the average user won't see any negative. So it'll help a lot if you can explain what reason remains to keep them

blakeembrey:
Honestly, arguments like this are a maintenance cost. Open issues are a maintenance cost. They’re the biggest maintenance cost I can possibly have. Nothing in the package is broken, but still I have to take hours out of my week to argue about things that shouldn’t need an argument. That is the cost, and I have to think about long term maintenance - I accept this and I might have another hundred PRs across things I maintain optimizing things that don’t matter. 

Do you understand how quickly these things add up? I accept one and then someone asks an AI to open it 100 times more, even if the benefit is negligible I now have to spend all my free time reviewing this stuff to make sure no one is trying to slip something by that could negatively impact the packages or my reputation.

I already stated, repeatedly, that I don’t have a good argument either way. 4kb over the wire and 30kb on disk is not a good argument. Neither is the marginal user that might benefit from cleaner source code during debugging.

For now I’m going to have to close this, I don’t have the bandwidth to engage faithfully. I said what I’d do, and that’s that for now.

blakeembrey:
Again, I want to be clear that doesn’t mean this won’t happen, but opening this PR and pushing so aggressively hasn’t felt like a good interaction. If it does occur, I’ll either reopen this PR or open it myself, that way I can ask the AI myself to open the hundred PRs and trust the result instead of having to review them all.

I still believe this is a risky road of optimization, but I respect it and could be on board if this could have been done differently. Setting an expectation of allowing dozens of PRs across the board to optimize package size is just asking for a future headache. And it has happened that something well meaning has lead me to feel burnout throughout periods of my OSS maintenance, so I’m sensitive to any potential fallout like that.

blakeembrey:
On a separate note, I'd suggest a change of tact. It's not a great feeling as a maintainer to be ganged up on by 4 people (3 comments, another +1'ing) and that has not endeared me to your cause. None of us are stupid, I think we can all appreciate the different sides of the argument and respect the opinions. In a single repo you are _probably_ correct, but I'm thinking about the other 200 repos I maintain and the next 10 years I'm stuck maintaining them, and from that perspective arguing about saving 30kb on a hard drive isn't something I want to be doing.

43081j:
for what it's worth, my comments were not intended to be "aggressive", I genuinely wanted to know if there's a reason I'm missing for wanting to keep the sourcemaps. That knowledge would help in future when deciding if to remove them from other packages too (including my own).

it is very unfortunate you saw this thread as a negative interaction. everyone who has replied here is trying to do good.

> I'm thinking about the other 200 repos I maintain and the next 10 years I'm stuck maintaining them, and from that perspective arguing about saving 30kb on a hard drive isn't something I want to be doing

I'm thinking about the thousands of repos in the ecosystem and the next 10 years that seemingly small decisions like this lead to petabytes of extra bandwidth and storage.

its a perspective thing. Nobody installs cookie by itself, every user installs it as part of a wider dependency tree. What seems insignificant to you (30KB disk, 4KB wire) adds up to a significant amount when other packages in that tree also add similar unused bytes.

On a side note, the rest of what you mention about burnout, AI, etc. Yes, everyone in this thread is a maintainer like you and unfortunately has to deal with the same things you're describing. It is a shame you have to deal with these things, and that attempts to do good can be held back by the fact that we're all overwhelmed these days.

thanks for your time anyway, hopefully we can revisit in future. Happy to try explain why these are unused bytes any time if the reasonings above didn't manage that.

PR #285, opened 2026-07-08
Perf optimize cookie value validation

blakeembrey:
Tweak encoder regex to be a negative class match (benchmark is slightly faster, but marginal, could undo if someone really disagrees) and skips validation of the value entirely when using the default encoder since it's already known valid.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -29,6 +29,11 @@ const cookieNameRegExp = /^[\u0021-\u003A\u003C\u003E-\u007E]+$/;
 const cookieValueRegExp = /^[\u0021-\u003A\u003C-\u007E]*$/;
 
 /**
+ * RegExp to match invalid cookie values to be encoded, including `%` for roundtrip safety.
+ */
+const cookieOctetInvalidRegExp = /[\u0000-\u0020\u007F-\uFFFF",;\\%]/;
+
+/**
  * RegExp to match domain-value in RFC 6265 sec 4.1.1
  *
  * domain-value      = <subdomain>
@@ -67,11 +72,6 @@ const pathValueRegExp = /^[\u0020-\u003A\u003D-\u007E]*$/;
  * RegExp to match max-age-value in RFC 6265 sec 5.6.2
  */
 const maxAgeRegExp = /^-?\d+$/;
-
-/**
- * RegExp to match RFC 6265 cookie-octet values (without % to preserve roundtrip) that need no URL encoding.
- */
-const cookieOctetRegExp = /^[!#$&'()*+\-.\/0-9:<=>?@A-Z[\]\^_`a-z{|}~]*$/;
 
 const NullObject = /* @__PURE__ */ (() => {
   const C = function () {};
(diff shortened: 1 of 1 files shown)


TeeAaTeeUu:
If more code is OK, an extra ~6% perf-increase can be achieved by replacing regex with a for-loop and individually checking `charCode`:

```typescript
function defaultEncode(str: string): string  {
  for (let i = 0, c = 0, len = str.length; i < len; i++) {
    c = str.charCodeAt(i);
    // \u0000-\u0020 (<33), \u007F-\uFFFF (>126), "(34), %(37), ,(44), ;(59), \(92)
    if (c < 33 || c > 126 || c === 34 || c === 37 || c === 44 || c === 59 || c === 92)
      return encodeURIComponent(str);
  }
  return str;
}
```

Runtime: `node 26.4.0 (arm64-darwin)`
case | before hz | after hz | Speedup |
| --- | ---: | ---: | ---: |
|empty | 25.380.759,34 | 25.288.506,03 | 1,00x |
|simple | 12.159.813,17 | 13.313.107,04 | 1,09x |
|rfc cookie-octets | 11.743.931,74 | 11.103.105,49 | 0,95x |
|encode | 5.753.640,44 | 6.171.864,26 | 1,07x |
|undefined values | 5.754.738,96 | 6.303.222,05 | 1,10x |
|mixed encode | 3.422.381,20 | 3.702.327,76 | 1,08x |
|10 cookies | 1.567.005,95 | 1.722.698,91 | 1,10x |
|100 cookies | 134.028,73 | 141.792,50 | 1,06x |

I think with these we are getting into the "_not so relevant speedup_" territory, meaning code-readability should be prioritised instead. If you read and understand regex more easily, then that is preferable than above for-loop!

TeeAaTeeUu:
And for completion, here also the other regexes in for-loop form, gaining ~10% more (_at least with the current bench, we might want to test with longer strings_):

```ts
function validCookieName(str: string): boolean  {
  for (let i = 0, c = 0, len = str.length; i < len; i++) {
    c = str.charCodeAt(i);
    // ;(59) =(61)
    if (c < 33 || c > 126 || c === 59 || c === 61)
      return false;
  }
  return true;
}

function validCookieValue(str: string): boolean  {
  for (let i = 0, c = 0, len = str.length; i < len; i++) {
    c = str.charCodeAt(i);
    // ;(59) 
    if (c < 33 || c > 126 || c === 59) return false;
  }
  return true;
}

function validCookiePath(str: string): boolean  {
  for (let i = 0, c = 0, len = str.length; i < len; i++) {
    c = str.charCodeAt(i);
    // ;(59) <(60)
    if (c < 32 || c > 126 || c === 59 || c === 60) return false;
  }
  return true;
}
```

And in case wanted to add a bit more readability, we can generate the `charCode` values:

```ts
const EXLAMATION = '!'.charCodeAt(0)
const DOUBLEQUOTE = '"'.charCodeAt(0)
const PERCENT = '%'.charCodeAt(0)
const COMMA = ','.charCodeAt(0)
const SEMICOLON = ';'.charCodeAt(0)
const LESS_THAN = '<'.charCodeAt(0)
const EQUAL = '='.charCodeAt(0)
const BACKSLASH = '\\'.charCodeAt(0)
const TILDE = '~'.charCodeAt(0)
```

TeeAaTeeUu:
Walking back on my suggestion...

For-loop perf improvement only happens when the invalid / encoding-required character is found within the first ~15 characters. By the time it's 512 chars, for-loop is twice as slow.

Meaning for-loop is still faster for longer strings, if for example `;` semicolon is found early within the string.

